PR #12, opened 2021-01-05
change default site_status to all

hellkite500:
The client defaults to fetching only "active" status gage data.  By default, the NWIS rest API uses "all" as the default parameter.  When retrieving historical data, this causes a bit of a surprise when a gage comes back with no data.

Closes #10 

## Changes

- Change default `site_status` from `active` to `all`

## Checklist

- [x] PR has an informative and human-readable title
- [x] PR is well outlined and documented. See [#12](https://github.com/jarq6c/evaluation_tools/pull/12) for an example
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output) using [numpy docstring](https://numpydoc.readthedocs.io/en/latest/format.html) formatting
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #41, opened 2021-02-17
Add optional parameter to filter events using minimum duration

jarq6c:
Introduces new optional parameter to filter events based on duration.

## Additions

- New `minimum_event_duration` option for `evaluation_tools.events.event_detection.decomposition` methods

## Removals

- None

## Changes

- Additional optional parameters and documentation

## Testing

1. Added `test_list_events_noise` method that simulates and tests a noisy signal

## Notes

- This is a common post-processing step. These algorithms are prone to false positives

## Todos

- None

## Checklist

### DIFF
--- a/python/events/evaluation_tools/events/event_detection/decomposition.py
+++ b/python/events/evaluation_tools/events/event_detection/decomposition.py
@@ -20,6 +20,8 @@ list_events
 import numpy as np
 import pandas as pd
 from typing import Union
+import datetime
+from pandas.tseries.frequencies import to_offset
 
 def detrend_streamflow(
     series: pd.Series,
@@ -145,7 +147,8 @@ def mark_event_flows(
 def list_events(
     series: pd.Series,
     halflife: Union[float, str, pd.Timedelta],
-    window: Union[int, pd.tseries.offsets.DateOffset, pd.Index]
+    window: Union[int, pd.tseries.offsets.DateOffset, pd.Index],
+    minimum_event_duration: Union[str, tuple, datetime.timedelta, pd.tseries.offsets.DateOffset] = '0H'
     ) -> pd.DataFrame:
     """Apply time series decomposition to mark event values in a streamflow
         time series. Discretize continuous event values into indiviual events.
@@ -163,6 +166,9 @@ def list_events(
         window: int, offset, or BaseIndexer subclass, required
             Size of the moving window for `pandas.Series.rolling.min`.
             This filter is used to model the trend in `series`.
+        minimum_event_duration: str, tuple, datetime.timedelta, pd.tseries.offsets.DateOffset, optional, default '0H'
+            Enforce a minimum event duration. This should generally be set equal to 
+            halflife to reduce the number of false positives flagged as events.
             
         Returns
         -------
@@ -185,8 +191,21 @@ def list_events(
     ends = ends[ends]
 
     # Extract times
-    return pd.DataFrame({
-        'start': starts.index,
-        'end': ends.index
-    })
+    if minimum_event_duration == '0H':
+        return pd.DataFrame({
+            'start': starts.index,
+            'end': ends.index
+        })
+    
+    # Build events DataFrame
+    events = pd.DataFrame({
+            'start': starts.index,
+            'end': ends.index
+        })
+
+    # Compute durations
+    durations = events['end'].sub(events['start'])
+
+    # Filter events
+    return events[durations >= to_offset(minimum_event_duration)].reset_index(drop=True)
     
--- a/python/events/evaluation_tools/events/event_detection/decomposition.py
+++ b/python/events/evaluation_tools/events/event_detection/decomposition.py
@@ -89,6 +89,39 @@ def detrend_streamflow(
     # Return detrended series
     return detrended
 
+def event_boundaries(event_points: pd.Series):
+    """Return a two column pandas.DataFrame with 'start' and 'end' event times
+        generated from a time series of boolean event points.
+        
+        Parameters
+        ----------
+        event_points: pandas.Series
+            Boolean time series where True indicates the associated value
+            in the `series` is part of an event.
+
+        Returns
+        -------
+        events: pandas.DataFrame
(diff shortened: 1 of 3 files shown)


python/events/evaluation_tools/events/event_detection/decomposition.py
```diff
@@ -139,13 +178,36 @@ def mark_event_flows(
     # Eliminate negative values
     detrended[detrended < 0.0] = 0.0
 
-    # Return mask of non-zero detrended flows
-    return pd.Series((detrended > 0.0), index=series.index)
+    # Do not filter events
+    if minimum_event_duration == '0H':
```
aaraney, line 182: You could remove the code redundancy by moving the conditional after `event_points` is constructed and just return `event_points` if the criteria is met. 

As for `minimum_event_duration == '0H':`, I think we may want to make this comparison more robust. So for example, the 0 hour `datetime.timedelta` is caught. Why im here, im probably just misunderstanding, but should one of the accepted types of `minimum_event_duration` be `pd.tseries.offsets.DateOffset`? I had not come across that object in `pandas` before myself and after a [little digging](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#overview) it looks like it is used to do calendar respective calculations, not temporally absolute (`pd.Timedelta`). Having played with a `pandas.tseries.offsets.DateOffset(hour=0)` , I can't find a way to easily get back a period string from the offset either just as a heads up. 

Also, a tuple as an excepted type? I am just a little confused. I checked that all of the object types you listed in the type hints can evaluate vs `==` without error, so that checks out. Just more curious than anything myself.

python/events/evaluation_tools/events/event_detection/decomposition.py
```diff
@@ -87,10 +89,44 @@ def detrend_streamflow(
...
+    starts = (event_points & ~forward_shift)
+    starts = starts[starts]
+
+    # Identify event ends
+    backward_shift = event_points.shift(-1).fillna(False)
+    ends = (event_points & ~backward_shift)
+    ends = ends[ends]
+
+    # Extract times
+    return pd.DataFrame({
```
aaraney, line 120: It seems like it is possible for `starts` and `ends` to be of different length. I am probably wrong about this, but is that true? If, so this should throw an error.

python/events/evaluation_tools/events/event_detection/decomposition.py
aaraney: Also having looked at how `event_points` is used later, namely `event_points.loc[:] = False`, it makes sense from the code readers perspective to change these names. Or just create an empty `pd.Series` with length `event_points`.

aaraney (changes requested):
Overall look great and im really glad to have this now. Its great that you can provide this argument as a period string. Mainly my comments are on type hints, handling one conditional, and repeated code.

python/events/evaluation_tools/events/event_detection/decomposition.py
```diff
@@ -177,23 +176,23 @@ def mark_event_flows(
...
+
     # Get list of potential events
     events = event_boundaries(event_points)
 
     # Compute durations
     durations = events['end'].sub(events['start'])
 
     # Filter events
-    events = events[durations >= to_offset(minimum_event_duration)].reset_index(drop=True)
+    events = events[durations >= pd.Timedelta(minimum_event_duration)].reset_index(drop=True)
```
aaraney, line 195: Still run into the issue with the `if minimum_event_duration == 0` conditional above. I think right before the conditional you should set `minimum_event_duration = pd.Timedelta(minimum_event_duration)`. Then the conditional should be `if minimum_event_duration == pd.Timedelta("0H")`

python/events/evaluation_tools/events/event_detection/decomposition.py
```diff
@@ -149,7 +148,7 @@ def mark_event_flows(
         window: int, offset, or BaseIndexer subclass, required
             Size of the moving window for `pandas.Series.rolling.min`.
             This filter is used to model the trend in `series`.
-        minimum_event_duration: str, tuple, datetime.timedelta, pd.tseries.offsets.DateOffset, optional, default '0H'
+        minimum_event_duration: pandas.Timedelta, datetime.timedelta, numpy.timedelta64, str, int, optional, default 0
```
aaraney, line 151: I liked what the default before, I think it was more intuitive. But that's a minor thing, either way.

aaraney (changes requested):
Just a minor change to a conditional.

python/events/evaluation_tools/events/event_detection/decomposition.py
jarq6c: I don't think this is possible. Both `starts` and `ends` are generated from the same `Series`. `shift` changes the index of the values, but does not alter the index. This is why both calls to `shift` are followed by `fillna`. The index is unchanged and the front and back of the new series have `NaN`.

python/events/evaluation_tools/events/event_detection/decomposition.py
jarq6c: OK. Thanks for this feedback. I changed `minimum_event_duration` to be a `pd.Timedelta` instead of an `offset`. This makes more sense. I also renamed the filtered series and generated a new series instead of re-assigning the values.

python/events/evaluation_tools/events/event_detection/decomposition.py
aaraney: Okay, that makes sense. Disregard my comment.

python/events/evaluation_tools/events/event_detection/decomposition.py
jarq6c: Changed back

python/events/evaluation_tools/events/event_detection/decomposition.py
jarq6c: Implemented.

aaraney:
Thanks for the fixes!

PR #42, opened 2021-02-18
Add option to shift event start times to local minimum

### LINKED ISSUE #37: Add option to shift start event times to local minima
Add a parameter that will shift start times produced by `evaluation_tools.events.event_detection.decomposition` to a local minima value.

jarq6c:
Phase shifts imparted by the constituent filters can cause event starts to miss the mark by a significant margin, especially on noisy data. Using this option can help fix that by shifting start times to a local minimum within some search window.

## Additions

- New `start_radius` option

## Removals

- None

## Changes

- Added new option and logic to `mark_event_flows`

## Testing

1. Verifies start is shifted to appropriate index for contrived noisy test event.

## Notes

- Closes #37 

## Todos

- None

## Checklist

### DIFF
--- a/python/events/evaluation_tools/events/event_detection/decomposition.py
+++ b/python/events/evaluation_tools/events/event_detection/decomposition.py
@@ -121,6 +121,34 @@ def event_boundaries(event_points: pd.Series):
         'end': ends.index
     })
 
+def find_local_minimum(origin, radius, timeseries):
+    """Return Datetime of local minimum value within radius of origin.
+        
+        Parameters
+        ----------
+        origin: pandas.Series index label applicable to timeseries, required
+            The original streamflow time series.
+        radius: pandas.Timedelta, datetime.timedelta, numpy.timedelta64, str, int, required
+            The search radius around origin to look for a minimum value.
+        timseries: pandas.Series with a DateTimeIndex, required
+            The original time series to inspect.
+            
+        Returns
+        -------
+        Index: 
+            Datetime of the minimum value.
+
+    """
+    # Define search radius
+    radius = pd.Timedelta(radius)
+
+    # Define window
+    left = origin - radius
+    right = origin + radius
+
+    # Find index of minimum value
+    return timeseries.loc[left:right].idxmin()
+
 def mark_event_flows(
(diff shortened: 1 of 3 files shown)


aaraney:
So just for clarification, the phase shift is due to the ewm smoothing step and the rolling minimum applied to the exponentially smoothed signal correct? So just as a rule of thumb the `start_radius` scale should be `<=` the ewm `halflife`? I would assume basin response rate would dictate the `start_radius` hyperparameter as well, given that a flashier basin should be picked up faster than a slower responding basin.

aaraney (approved):
All of this looks great.

jarq6c:
> So just for clarification, the phase shift is due to the ewm smoothing step and the rolling minimum applied to the exponentially smoothed signal correct? So just as a rule of thumb the `start_radius` scale should be `<=` the ewm `halflife`? I would assume basin response rate would dictate the `start_radius` hyperparameter as well, given that a flashier basin should be picked up faster than a slower responding basin.

Yes, setting `start_radius <= halflife` is probably a good place to start. The phase shift is probably mostly attributable to the noise cutoff. The EWMA imparts some shift, but not too much for relatively small `halflife`. The way the minimum filters are implemented theoretically results in 0 shift. However, they still result in attenuation of the signal which can push the starts forward in time after thresholding for noise. This is because the early event rise is likely to fall within the noise threshold, especially after the original signal is smoothed.

The current noise threshold assumes the detrended signal has a zero median. This assumption is reasonable if events are sufficiently rare. So what ever residual remains after detrending must be noise. Subtracting the detrended median from every value removes the negative fluctuations of noise. Subtracting twice the detrended median removes both the positive and negative fluctuations. Imagine a sine wave oscillating around the zero line from positive to negative. Unfortunately, thresholding in this way clips off the bottom of events. This behavior is good for event ends, but not starts.

We have yet to rigorously explore the behavior of this filter. I suspect the most appropriate `start_radius` will likely be a function of signal-to-noise ratio and flashiness (as you surmised).

jarq6c:
> All of this looks great.

Thanks for the review!

PR #44, opened 2021-02-19
Raise exceptions for bad time series index in event detection

jarq6c:
Adds additional code to validate time series and test exceptions.

## Additions

- Tests for non-duplicated monotonically increasing datetime index

## Removals

- None

## Changes

- `mark_event_flows` and `find_local_minimum` raise exceptions for invalid series

## Testing

1. New tests verify exceptions are raised for bad series

## Notes

- Lower level methods will also catch some of these, but these exceptions make them explicit

## Todos

- None

## Checklist

### DIFF
--- a/python/events/evaluation_tools/events/event_detection/decomposition.py
+++ b/python/events/evaluation_tools/events/event_detection/decomposition.py
@@ -55,6 +55,16 @@ def detrend_streamflow(
             New streamflow time series with trend removed.
         
         """
+    # Check index
+    if type(series.index) != pd.DatetimeIndex:
+        raise Exception("series index is not DatetimeIndex")
+
+    if not series.index.is_monotonic_increasing:
+        raise Exception("series index is not monotonically increasing")
+
+    if series.index.has_duplicates:
+        raise Exception("series index has duplicate timestamps")
+
     # Flip series values to run filter in reverse
     if reverse:
         series = pd.Series(series.values[::-1], 
--- a/python/events/setup.py
+++ b/python/events/setup.py
@@ -13,7 +13,7 @@ SUBPACKAGE_NAME = "events"
 SUBPACKAGE_SLUG = f"{NAMESPACE_PACKAGE_NAME}.{SUBPACKAGE_NAME}"
 
 # Subpackage version
-VERSION = "0.2.3"
+VERSION = "0.2.4"
 
 # Package author information
 AUTHOR = "Jason Regina"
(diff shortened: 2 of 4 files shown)


aaraney:
Hmmm, so in the context of "when to raise an exception", I am not sure that all of these merit an exception. I agree with the index not being a `pd.DatetimeIndex` this is cause things to break. But in the other cases, if you are to pass a series that has a datetime index, however is not monotonically increasing and/or has duplicate timestamps, will it cause an exception prior to this PR? I am just wondering if these are more appropriate as raised warnings instead of exception. 

I bring it up mainly in the context of someone who uses this in a workflow, should the rest of the workflow fail (even if it is not related to event detection) if there are duplicates or the datetime index is not monotonically increasing? I think this is up for debate, but wanted to mention it.

jarq6c:
> Hmmm, so in the context of "when to raise an exception", I am not sure that all of these merit an exception. I agree with the index not being a `pd.DatetimeIndex` this is cause things to break. But in the other cases, if you are to pass a series that has a datetime index, however is not monotonically increasing and/or has duplicate timestamps, will it cause an exception prior to this PR? I am just wondering if these are more appropriate as raised warnings instead of exception.
> 
> I bring it up mainly in the context of someone who uses this in a workflow, should the rest of the workflow fail (even if it is not related to event detection) if there are duplicates or the datetime index is not monotonically increasing? I think this is up for debate, but wanted to mention it.

As a check, I tested these scenarios outside of `evaluation_tools`.

The specific method where I implemented the checks is `detrend_streamflow`.  The `ewm` and `rolling` methods will fail if the index is not `DatetimeIndex` and not monotonic. I think we have to minimally raise for these cases. We can probably safely warn about the duplicated index for now. However, there are methods we might add later that will fail if the index is duplicated. Duplicated indexes are also likely to happen because of how NWIS returns data.

There may be a need for a convenience `preprocess` method a user can employ to prepare their time series before running event detection.

aaraney:
> As a check, I tested these scenarios outside of `evaluation_tools`.

Thanks for looking into these edge cases. From your findings, it sounds like its best to move forward how you've proposed here. I think it safe to leave the exception for duplicate datetimes as a common sense thing. 

> There may be a need for a convenience `preprocess` method a user can employ to prepare their time series before running event detection.

Yeah that is a possibility. I wonder how difficult it would be to create a dataframe validation/preprocessing method that is more generic. So you pass it a dataframe and it verifies that it is a `evaluation_tools` compliant dataframe and or spits resolves any datatype issues. I could see this sort of method fitting well in a `utilities` subpackage. Just a thought.

aaraney (approved):

PR #45, opened 2021-02-22
Light Event Detection Refactor

jarq6c:
This PR cleans up some of the event detection code. It reduces the number of computations and separates the trend computation step from the detrend step more explicitly. This model will make more sense as we add more trend estimation techniques.

## Additions

- `rolling_minimum` function is the method used to estimate the trend.

## Removals

- Function parameters were removed where applicable.

## Changes

- I found a bug in the method that tests event detection on noisy series. The scale and frequency of the noise was incorrect. For some reason I simulated ~1 fluctuation a day, but given the parameters, I clearly meant ~4 fluctuations a day. Also found better results by setting `start_radius` to 1 period over the expected noise frequency (`halflife`). This makes sense since the period of the simulated noise is actually slightly more than `'6H'`.

## Testing

1. Passes all other tests with correction mentioned above.

## Notes

- Nothing in the main entry functions `list_events` and `mark_event_flows` has changed. So nothing downstream should break.

## Todos

- None

## Checklist

### DIFF
--- a/python/events/evaluation_tools/events/event_detection/decomposition.py
+++ b/python/events/evaluation_tools/events/event_detection/decomposition.py
@@ -47,6 +47,10 @@ def rolling_minimum(
             Approximate baseflow trend of original series.
         
         """
+    # Check index
+    if not isinstance(series.index, pd.DatetimeIndex):
+        raise Exception("series index is not DatetimeIndex")
+
     # Estimate trend using a forward rolling minimum
     forward = series.rolling(window=window).min()
 
--- a/python/events/evaluation_tools/events/event_detection/decomposition.py
+++ b/python/events/evaluation_tools/events/event_detection/decomposition.py
@@ -11,8 +11,11 @@ Regina, J.A., F.L. Ogden, 2020.  Automated Correction of Systematic
 
 Functions
 ---------
+rolling_minimum
 detrend_streamflow
 mark_event_flows
+find_local_minimum
+event_boundaries
 list_events
 
 """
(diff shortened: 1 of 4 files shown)


python/events/evaluation_tools/events/event_detection/decomposition.py
```diff
@@ -22,11 +25,49 @@
...
+        window: int, offset, or BaseIndexer subclass, required
+            Size of the moving window for `pandas.Series.rolling.min`.
+            This filter is used to model the trend in `series`.
+            
+        Returns
+        -------
+        trend: pandas.Series
+            Approximate baseflow trend of original series.
+        
+        """
```
aaraney, line 49: It seems appropriate to check if the index of the series is a `pd.DatetimeIndex`. This was done in `detrend_streamflow`. So something like:


```python
if isinstance(series.index, pd.DatetimeIndex) is False:
  # Throw an exception
```

aaraney (changes requested):
Really minor change. Overall looks great!

Interesting your note about the `start_radius` parameter, that makes sense, i'm just curious how we can better estimate the `dt`, the `dt` being how the difference between the `halflife` parameter and the "intended result". Just kind of spitballing, I wonder if `dt` is proportional to the slope change in the `ewma`. Therefore, `dt` would be something like `e^ (something)`. Not really forming a complete idea here, just more or less thinking out loud.

python/events/evaluation_tools/events/event_detection/decomposition.py
aaraney: 7af6524 addresses this.

aaraney (approved):

python/events/evaluation_tools/events/event_detection/decomposition.py
jarq6c: Is there a conventional or standard way to check for this?

Examples:
```python
if isinstance(series.index, pd.DatetimeIndex) is False:
  # Throw an exception

if not isinstance(series.index, pd.DatetimeIndex):
  # Throw an exception

if type(series.index) != pd.DatetimeIndex:
  # Throw an exception
```

Issue #51, opened 2021-03-17
Evaluate the influence of NaN values on event detection


jarq6c:
Possible bug. The event detection methods validate the index, but not the values. The presence of NaN in the value series may produce undefined behavior given the recursive nature of the filters. I want to see if this case needs handling.

jarq6c:
The filters seem robust again NaN values. The algorithm completes and returns data if NaNs are present. I ran the example below using a 7-day (larger than the gap) and a 30-minute (smaller than the gap) window. The NaNs will definitely influence the deviation from the median signal and effect where the start and end event time land. It's still not recommended to run event detection on time series with missing data, but it won't shut down the process and at least in this case return consistent (if not sensible) "events."

```python
# Import tools to retrieve data and detect events
from hydrotools.nwis_client.iv import IVDataService
from hydrotools.events.event_detection import decomposition as ev
import matplotlib.pyplot as plt

# Retrieve streamflow observations for two sites
service = IVDataService(
    value_time_label="value_time"
    )
observations = service.get(
    sites='01360640', 
    startDT='2021-07-01', 
    endDT='2021-08-01'
    )

# Drop extra columns to be more efficient
observations = observations[[
    'value_time', 
    'value'
    ]]

# Check for duplicate time series, keep first by default
observations = observations.drop_duplicates(
    subset=['value_time']
    )

# Resample to hourly, keep first measurement in each 1-hour bin
observations = observations.set_index("value_time")
observations = observations.resample("15min").nearest(limit=1)

# Detect events
events = ev.list_events(
    observations['value'],
    halflife='6H', 
    window='7D',
    start_radius="24H"
    )

# Print event list    
print(events)

# Plot
observations.plot(logy=True)
observations.loc[events.start, "value"].plot(style="o", ax=plt.gca())
observations.loc[events.end, "value"].plot(style="o", ax=plt.gca())
plt.show()
```
![day7](https://user-images.githubusercontent.com/12818140/128706772-d5d91362-670c-4416-a375-db32df4ebc54.png)
![minute30](https://user-images.githubusercontent.com/12818140/128706774-7312c091-d4ea-4397-8c2b-c7029d382b0c.png)

jarq6c:
I do think missing values ought to be handled in a preprocessing step and the README/docs recommend tips to run event detection, including filling missing data. However, this burden is on the caller. There are too many types of signals and data to account for all the possible ways a given time series could be filled. In some cases, it comes down to personal preference.

[...]

PR #52, opened 2021-03-17
Ensure filters aren't run on missing data

hellkite500:
Related to #51, here is one way of potentially handling inputs that do contain NaN.  This makes no interpolation assumptions, but simply only marks events continuous chunks of the input series.

## Additions

- Add handling of NaN in `list_events` input series

## Testing

1. All existing tests pass
2. An additional test with missing data should probably be added

## Todos

- Add  test case for series with missing data

## Checklist

- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output) using [numpy docstring](https://numpydoc.readthedocs.io/en/latest/format.html) formatting

### DIFF
--- a/python/events/evaluation_tools/events/event_detection/decomposition.py
+++ b/python/events/evaluation_tools/events/event_detection/decomposition.py
@@ -322,8 +322,17 @@ def list_events(
             `end` columns indicate the boundaries of each event.
         
         """
+    missing = series.isnull()
+    #Find all groups of consecutive data
+    #True = 1, False = 0, so a cumulative sum
+    #identifies groups of data (falses)
+    grouper = missing.cumsum()
+    #Drop groups of nan
+    grouper = grouper[ series.notnull() ]
+    #Group the original data based on this cumulative sum
+    groups = series.groupby(grouper)
     # Detect event flows
-    event_points = mark_event_flows(series, halflife, window, 
+    event_points = groups.apply( mark_event_flows, halflife, window,
         minimum_event_duration, start_radius)
 
     # Return events


jarq6c:
@hellkite500 My preference to reject to discontinuous series and let users either fill or implement this type of groupby logic outside the list events methods.

hellkite500:
A note on pandas groupby/apply -- it isn't guaranteed to always return the same data type.  In some cases, it is possible for the apply to return a `Series` where in other's it may return a `Dataframe`.  This largely depends on the input and the resulting set of outputs from the input.  But that means to use it in this setting, one would need to check if the `event_points` is `empty` (in which case `event_points` would be a `Series`) or not (in which case it will be the expected `DataFrame`)

jarq6c:
After further conversation, I think we should warn about missing data and offer this functionality through some type of `evaluation_tools.preprocessing` module.

My thoughts:
1. The underlying filters won't break if NaN is present
2. Dealing with missing data is a science. I'm afraid integrating any default handling of missing data could be perceived as an endorsement of that method
3. I think we should avoid embedding preprocessing into event detection as much as possible (this version of event detection actually purged a lot of preprocessing)
4. This particular method of preprocessing/NaN handling has its own issues that I think can be more cleanly tested and maintained in a separate module
5. Raising a warning about NaN is any easy compromise to avoid "surprises"

In particular, I think this method will need to somehow check that the length of each continuous portion makes sense given the parameters (`window` and `halflife`). As near as I can tell the method does not distinguish between gaps and a degraded signal. It is not uncommon for missing data to occur during high-flow events, this method might split those events. This method has its own set of parameters that could be tweaked depending upon the nature of the underlying signal and the events you're trying to detect. This method does not seem any less opinionated in its treatment of NaN than other methods of no-data handling.

python/events/evaluation_tools/events/event_detection/decomposition.py
```diff
@@ -322,8 +322,17 @@ def list_events(
             `end` columns indicate the boundaries of each event.
         
         """
+    missing = series.isnull()
```
jarq6c, line 325: I think we could raise a warning here:

```python
if series.isnull().any():
    raise
```

python/events/evaluation_tools/events/event_detection/decomposition.py
```diff
@@ -322,8 +322,17 @@ def list_events(
...
+    #True = 1, False = 0, so a cumulative sum
+    #identifies groups of data (falses)
+    grouper = missing.cumsum()
+    #Drop groups of nan
+    grouper = grouper[ series.notnull() ]
+    #Group the original data based on this cumulative sum
+    groups = series.groupby(grouper)
     # Detect event flows
-    event_points = mark_event_flows(series, halflife, window, 
+    event_points = groups.apply( mark_event_flows, halflife, window,
```
jarq6c, line 335: I think it makes more sense to implement this in a preprocessing module. So the preprocessing splits a series with gaps into multiple series and then event event detection is ran using `groupby` on those subseries. This would allow users to tweak how the series is split perhaps with an option to perform limited filling.

jarq6c (commented):
I think this feature has value, but is outside scope of the event detection module.

hellkite500:
> 1. The underlying filters won't break if NaN is present
What about the case where NaN causes an  "event" of all 0  flow's  to be identified?  Is  the an artifact of the filter's neglecting the missing data?   What does it mean to "break"  the filters haha?

> 2. Dealing with missing data is a science. I'm afraid integrating any default handling of missing data could be perceived as an endorsement of that method
I  don't disagree.  Providing a `preprocessing` module isn't a bad idea.  I still think it is worth integrating that module via kwargs or  function overloads to present the user a  simple calling sequence to get events.

> 3. I think we should avoid embedding preprocessing into event detection as much as possible (this version of event detection actually purged a lot of preprocessing)
As per the previous comment, would it  make sense to provide some more tightly integrated options to simplify the calling stack?

> 4. This particular method of preprocessing/NaN handling has its own issues that I think can be more cleanly tested and maintained in a separate module
Agreed.

> 5. Raising a warning about NaN is any easy compromise to avoid "surprises"
A warning or an actual exception/error?

jarq6c:
Zero flow events resulting from NaN are caused by the NaN in the original signal, not the event detection algorithm directly. By "break", I meant a NaN won't halt the entire process. Event detection and the underlying filters generate models of time series components. As with all models, garbage-in begets garbage-out.

I think simplifying the calling stack is a matter of personal preference worth qualifying. Which of these would you consider "simpler"?
```python
# Example A
events = list_events(series, halflife, window, no_data_method, keyword1, keyword2)

# Example B
preprocessed_series = no_data_method(series, keyword1, keyword2)
events = preprocessed_series.groupby('subseries').apply(list_events, halflife, window)
```
I think Example A's interface seems simpler, but the code to handle `no_data_method` inside of `list_events` may be quite complex, especially if we're going to support multiple methods to handle NaNs. I think this will be difficult to test and maintain in the event detection module. I think Example B offers more utility and will be easier to test and maintain.

Another way to offer a more tightly integrated calling stack is through `evaluation_tools.workflows`. These are pre-developed data-driven computational workflows with prescribed steps for data retrieval, preprocessing, processing, analysis, and visualization. The goal here is to model the workflow as a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph).
```python
# Example C
run_my_event_based_analysis(observations='NWIS', simulations='NWMv2.1', date_range=('1970-01-01','2000-01-01'),
    output='output_directory')
```

I vote for warning about NaN, but not raising an exception. We may want to raise an exception if a significant portion of the series is all NaN.

jarq6c:
Will revisit in a different part of the `events` subpackage.

PR #67, opened 2021-04-21
_restclient utilities module: add Alias and AliasGroup classes

aaraney:
## Additions

- `utilities` module under `_restclient` subpackage
- Immutable `Alias` class that accepts a key of any type and a value or collection of values that act as the alias to the key. To get key, use `.get(value) # returns key or None` or `[value] # throws ValueError if invalid value`.
- Immutable `AliasGroup` that supports a collection of `Alias` objects but treats them as a single `Alias`.
- `IterableNonStringLike` type for checking if object is an instance of an iterable type but not a string or byte. string and bytes are iterable in python3.

PR adds `hydrotools._restclient.utilities` which contains: Immutable `Alias` and `AliasGroup` classes. These classes serve to simplify library code where it is common to separate back-end implementation from the front-end api. Often this is applicable when developing a library to access a rest api. 

An concrete use-case for this feature is interfacing with the [USGS NLDI](https://waterdata.usgs.gov/blog/nldi-intro/) [REST api](https://labs.waterdata.usgs.gov/api/nldi/swagger-ui/index.html?configUrl=/api/nldi/v3/api-docs/swagger-config). The NLDI api provides endpoints for accessing to stream and basin features on and along the NHDPlus network from a variety of location metadata providers (USGS NWIS, NHDPlus comid, HUC12 Pour Points, etc.). The api is configured as follows:

```shell
# `location-provider` and `location` are variables to the api
https://labs.waterdata.usgs.gov/api/nldi/linked-data/{{location-provider}}/{{location}}/navigation/
```

Some valid `location-providers` are:
`comid`: NHDplus comid
`nwissite`: USGS NWIS
`huc12pp`: HUC12 Pour Points

As a developer, instead of these "back-end" provider names I would rather my API refer to these `location-providers` as `nhd`, `usgs` or `nwis`, and `huc12`. This could be accomplished in several ways, but this feature addition simplifies the desired behavior.

```python
comid = Alias("comid", "nhd")
nwissite = Alias("nwissite", ["usgs", "nwis"])
huc12pp = Alias("huc12pp", "huc12")

location_provider_group = AliasGroup([comid, nwissite, huc12pp])

# Usage
location_provider_group.get("nhd") # returns "comid"

# Or using bracket notation
location_provider_group["nwis"] # returns "nwissite"

nwissite["usgs"] # returns "nwissite"
```

**All** types are valid as keys, meaning `Alias` could also be used in a factory pattern where the key is a function.

## Testing

1. `Alias` and `AliasGroup` tests added
2. `IterableNonStringLike` tests added

## Checklist

### DIFF
--- a/python/_restclient/hydrotools/_restclient/_iterable_nonstring.py
+++ b/python/_restclient/hydrotools/_restclient/_iterable_nonstring.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+from abc import ABCMeta, abstractmethod
+
+__all__ = ["IterableNonStringLike"]
+
+
+class IterableNonStringLike(metaclass=ABCMeta):
+    """Type iterable that is not a subclass of str or bytes. Useful to test if object
+    `isinstance()`.
+
+    Example:
+        x = "a-string"
+        y = list()
+        isinstance(x, IterableNonStringLike) # False
+        isinstance(y, IterableNonStringLike) # True
+    """
+
+    slots = ()
+
+    @abstractmethod
+    def __iter__(self):
+        while False:
+            yield None
+
+    @classmethod
+    def __subclasshook__(cls, C):
+        if cls is IterableNonStringLike:
+            if any(
+                "__iter__" in B.__dict__
+                and not issubclass(B, str)
+                and not issubclass(B, bytes)
+                for B in C.__mro__
+            ):
+                return True
+        return NotImplemented
--- a/python/_restclient/hydrotools/_restclient/utilities.py
+++ b/python/_restclient/hydrotools/_restclient/utilities.py
@@ -219,9 +219,9 @@ class AliasGroup:
         Returns
         -------
         Union[int, float, str, bytes, bool, Callable, tuple, frozenset, None]
-           alias value if valid value, else None
+           alias value if valid key, value, or Alias instance, else None
         """
-        option = self.option_map.get(value)
+        option = self.option_map.get(key)
 
         if option is None:
             return None
@@ -230,15 +230,20 @@ class AliasGroup:
 
     @property
     def option_groups(self):
+        """ Mapping of Alias keys's to Alias value's """
         return self._option_groups
 
-    def __getitem__(self, value):
-        option = self.get(value)
+    @property
+    def values(self):
+        """ Frozenset of present Alias value's """
+        return self._values
+
+    def __getitem__(self, key):
+        option = self.get(key)
 
         if option is None:
             raise ValueError(
-                "Invalid value %s. Valid values are %s"
-                % (value, self.option_map.keys())
+                f"Invalid value {str(key)}. Valid values are {str(self.option_map.keys())}"
             )
 
         return option
--- a/python/_restclient/hydrotools/_restclient/utilities.py
+++ b/python/_restclient/hydrotools/_restclient/utilities.py
@@ -234,9 +234,14 @@ class AliasGroup:
         return self._option_groups
 
     @property
-    def values(self):
+    def values(self) -> frozenset:
         """ Frozenset of present Alias value's """
         return self._values
+
+    @property
+    def keys(self) -> set:
+        """ Alias keys """
+        return set(self.option_map.keys())
 
     def __getitem__(self, key):
         option = self.get(key)
--- a/python/_restclient/hydrotools/_restclient/utilities.py
+++ b/python/_restclient/hydrotools/_restclient/utilities.py
@@ -51,13 +51,24 @@ class Alias:
             self.__dict__["valid_value"] = self.valid_value | self.key.valid_value
             self.__dict__["key"] = self.key.key
 
-    def get(self, value) -> Any:
+    def get(self, value: Hashable) -> Union[Any, None]:
+        """Get key given a valid alias value. If a valid key is not provided, return
+        None.
+
+        Parameters
+        ----------
+        value : Hashable
+            Valid alias value
+
+        Returns
+        -------
+        Union[Any, None]
+           alias key if valid value, else None
+        """
         if value in self:
             return self.key
 
-        raise ValueError(
-            "Invalid value %s. Valid values are %s" % (value, self.valid_value)
-        )
+        return None
 
     def __contains__(self, value) -> bool:
         return value in self.valid_value
--- a/python/_restclient/hydrotools/_restclient/utilities.py
+++ b/python/_restclient/hydrotools/_restclient/utilities.py
@@ -32,12 +32,12 @@ class Alias:
         result = foo_alias["bar"]()
     """
 
-    key: Any
+    value: Any
     valid_value: Union[Iterable, Hashable]
 
     def __post_init__(self):
-        # Create deep copy so a ref to a mutable key could not change value implicitly
-        self.__dict__["key"] = deepcopy(self.key)
+        # Create deep copy so a ref to a mutable value could not change keys implicitly
+        self.__dict__["value"] = deepcopy(self.value)
 
         # If non- str/bytes collection, frozenset, else frozenset([valid_value])
         self.__dict__["valid_value"] = (
@@ -46,13 +46,13 @@ class Alias:
             else frozenset([self.valid_value])
         )
 
-        if isinstance(self.key, Alias):
-            # Alias is passed as key, use contents to extend into new instance
-            self.__dict__["valid_value"] = self.valid_value | self.key.valid_value
-            self.__dict__["key"] = self.key.key
+        if isinstance(self.value, Alias):
+            # Alias is passed as value, use contents to extend into new instance
+            self.__dict__["valid_value"] = self.valid_value | self.value.valid_value
+            self.__dict__["value"] = self.value.value
 
     def get(self, value: Hashable) -> Union[Any, None]:
-        """Get key given a valid alias value. If a valid key is not provided, return
+        """Get value given a valid alias value. If a valid value is not provided, return
         None.
 
         Parameters
@@ -63,10 +63,10 @@ class Alias:
         Returns
         -------
         Union[Any, None]
-           alias key if valid value, else None
+           alias value if valid value, else None
         """
         if value in self:
-            return self.key
+            return self.value
 
         return None
 
@@ -74,20 +74,20 @@ class Alias:
         return value in self.valid_value
 
     def __getitem__(self, value) -> Any:
-        key = self.get(value)
+        value = self.get(value)
 
-        if key is None:
+        if value is None:
             raise ValueError(
                 "Invalid value %s. Valid values are %s" % (value, self.valid_value)
             )
 
-        return key
+        return value
 
     def __str__(self):
-        return str(self.key)
+        return str(self.value)
 
     def __repr__(self):
-        return f"{str(self.key)}: {str(self.valid_value)}"
+        return f"{str(self.value)}: {str(self.valid_value)}"
 
 
 class AliasGroup:
@@ -144,8 +144,8 @@ class AliasGroup:
             raise ValueError(f"Repeated valid_value {duplicate_value} not allowed")
 
     def get(self, value: Hashable) -> Union[Any, None]:
-        """Get singular Alias key from group when provided valid alias value. If a
-        valid key is not provided, return None.
+        """Get singular Alias value from group when provided valid alias value. If a
+        valid value is not provided, return None.
 
         Parameters
         ----------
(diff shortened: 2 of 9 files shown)


jarq6c (changes requested):
Looks generally fine to my eyes. I found three issues you might want to resolve:

What is the expected behavior with `dict` which is an iterable?
```python
from hydrotools._restclient.utilities import Alias

velocity_units = {
    'meters_per_second': 'm/s',
    'feet_per_second': 'ft/s'
}

velocity = Alias('velocity', velocity_units)
````

It seems `key` is not an alias for itself:

```python
cms = Alias("cms", ["CMS", "m^3/s"])
print("cms" in cms) # returns False
```

Have you considered an aliases property and or an iterable way to list aliases?
```python
johns_names = Alias("John Smith", ["Johnny Boy", "Papa John"])

# Perhaps list all aliases
print(johns_names.aliases)

# ...or iterate over aliases
for alias in johns_names:
    print(alias)
```

aaraney:
Thanks for having a look at this. My inline responses are below.

> What is the expected behavior with `dict` which is an iterable?
> 
> ```python
> from hydrotools._restclient.utilities import Alias
> 
> velocity_units = {
>     'meters_per_second': 'm/s',
>     'feet_per_second': 'ft/s'
> }
> 
> velocity = Alias('velocity', velocity_units)
> ```

They keys of the passed `values` dictionary should be the alias to the key passed to the `Alias` constructor.


> It seems `key` is not an alias for itself:
> 
> ```python
> cms = Alias("cms", ["CMS", "m^3/s"])
> print("cms" in cms) # returns False
> ```

Yeah that's a fair point. My only concern is breaking the ability to alias a function or class as the key. So, `Alias(lambda x: x**2, ["square"])`. Since:

```python
x = lambda x: x**2
y = lambda x: x**2

x is y # Returns False
x == y # Returns False
```

The point is well taken though. I am a little weary given that you are implicitly assigning an alias (to itself), but again im probably not being so pragmatic here. Thoughts?

> Have you considered an aliases property and or an iterable way to list aliases?
> 
> ```python
> johns_names = Alias("John Smith", ["Johnny Boy", "Papa John"])
> 
> # Perhaps list all aliases
> print(johns_names.aliases)
> 
> # ...or iterate over aliases
> for alias in johns_names:
>     print(alias)
> ```

I'd not thought to do that. That makes perfect sense to be able to do both of those actions. Ill add the functionality.

jarq6c:
It may pay to think about what exactly is being aliased here and what your intentions are for this improvement.

Are `dict` keys becoming aliases consistent with your design intent?

Given how memory management works in Python, what exactly is an "Alias"? Testing the equivalence of functions is a little odd. Since logical operators will fail in this case, I would imagine checking for the presence of a function or class alias would also fail unless you only compare specific instances (or byte code, which is prone to its own errors). You may need to implement overrides for comparison operators on `Alias`.

Are Aliases alternative labels for things or alternative variables for the same thing?

How do you expect Aliases to handle the following situation?
```python
# Situation A
x = lambda z: z **2
y = x
print(y == x) # True

# Situation B
x = lambda z: z**2
y = lambda z: z**2
print(y == x) # False
```

python/_restclient/hydrotools/_restclient/utilities.py
```diff
@@ -0,0 +1,181 @@
...
+
+        result = foo_alias["bar"]()
+    """
+
+    key: Any
+    valid_value: Union[Iterable, Hashable]
+
+    def __post_init__(self):
+        # Create deep copy so a ref to a mutable key could not change value implicitly
+        self.__dict__["key"] = deepcopy(self.key)
```
hellkite500, line 40: The assumption that `Any` type is `deepcopy`able may not be appropriate.  Especially for custom and/or extension types.

python/_restclient/hydrotools/_restclient/utilities.py
```diff
@@ -0,0 +1,181 @@
...
+                "Invalid value %s. Valid values are %s" % (value, self.valid_value)
+            )
+
+        return key
+
+    def __str__(self):
+        return str(self.key)
+
+    def __repr__(self):
+        return f"{str(self.key)}: {str(self.valid_value)}"
```
hellkite500, line 90: Pick a string formatting technique and use it throughout.  Here you use `f""` style format strings, in the exception above you use `%` style.

python/_restclient/hydrotools/_restclient/utilities.py
```diff
@@ -0,0 +1,181 @@
...
+            symmetric_differences = member.valid_value ^ symmetric_differences
+            union = member.valid_value | union
+
+            for valid_value in member.valid_value:
+                self.option_map[valid_value] = member
+
+        duplicate_value = union - symmetric_differences
+
+        if duplicate_value:
+            raise ValueError(f"Repeated valid_value {duplicate_value} not allowed")
```
hellkite500, line 139: This is not intuitive, since this is on the alias values of the passed in alias features.  The messaging and doc string should make this distinction better.

python/_restclient/tests/test_utilities.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_cannot_alter_alias_key(alias_fixture):
+    from dataclasses import FrozenInstanceError
+
+    with pytest.raises(FrozenInstanceError):
+        alias_fixture.key = "that"
+
+
+def test_cannot_alter_alias_key(alias_fixture):
```
hellkite500, line 34: duplicate of previous test

hellkite500: Actually, name is duplicated, function is not.  This actually tests changing the value, not the key.  Rename to `test_cannot_alter_alias_value`

python/_restclient/hydrotools/_restclient/utilities.py
```diff
@@ -0,0 +1,181 @@
...
+
+            url = f"{base_url}/{path}/cool-feature"
+
+            response = requests.get(url)
+            return response.json()
+    """
+
+    def __init__(self, alias: List[Alias]) -> None:
+
+        self._option_groups = (
```
hellkite500, line 120: Guarantee that only Alias types are passed in `alias` or you won't be able to maintain your immutability assumption in the group.

python/_restclient/hydrotools/_restclient/_iterable_nonstring.py
```diff
@@ -0,0 +1,35 @@
...
+        y = list()
+        isinstance(x, IterableNonStringLike) # False
+        isinstance(y, IterableNonStringLike) # True
+    """
+
+    slots = ()
+
+    @abstractmethod
+    def __iter__(self):
+        while False:
```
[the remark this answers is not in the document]
aaraney: I validated that your suggestion works with the tests i'd written, I was just following the python `_collection_abc.py` implimentation of `Iterable`:

```python
class Iterable(metaclass=ABCMeta):

    __slots__ = ()

    @abstractmethod
    def __iter__(self):
        while False:
            yield None

    @classmethod
    def __subclasshook__(cls, C):
        if cls is Iterable:
            return _check_methods(C, "__iter__")
        return NotImplemented
```

Im not sure if my tests cover every case so I am just going to stick to how its implemented in the `_abc` source.

python/_restclient/hydrotools/_restclient/_iterable_nonstring.py
```diff
@@ -0,0 +1,35 @@
...
+            yield None
+
+    @classmethod
+    def __subclasshook__(cls, C):
+        if cls is IterableNonStringLike:
+            if any(
+                "__iter__" in B.__dict__
```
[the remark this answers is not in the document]
aaraney: I get what you mean, but just to follow suit with how python's `typing` module treats "iterable" objects im going to leave this as is.

```python
class ImplicitlyIterable:
    def __getitem__(self, item):
        pass


def test_is_iterable():
    from typing import Iterable

    assert isinstance(ImplicitlyIterable, Iterable) # False
```

python/_restclient/hydrotools/_restclient/utilities.py
aaraney: Yeah having read this again I agree haha. Ill update the message to make it more clear.

python/_restclient/hydrotools/_restclient/utilities.py
aaraney: Agreed, having done some reading to see if I could figure out a way to determine if an object will be `deepcopy`able at construction and [this](https://stackoverflow.com/a/18611983) is all that ive found. It looks like an extension type would have to be pickable for it to be `deepcopy`able. Unfortunately I havent found an easy way to check if an object is pickable or not without just pickling the object itself. This seems like unnecessary overhead, so as we talked about offline I think the best course of action is to only accept Callables or Scalar types here.

python/_restclient/tests/test_utilities.py
aaraney: Nice catch!

python/_restclient/hydrotools/_restclient/utilities.py
aaraney: Yeah I agree that being stricter here about the accepted input types is probably a good idea.

jarq6c (approved):

python/_restclient/hydrotools/_restclient/_iterable_nonstring.py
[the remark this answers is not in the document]
hellkite500: Well `while False` will never invoke the `yield` so this function always does nothing (returns None) which is the same as `pass`.  Its just odd form.

aaraney:
In light of the above comments and discussion with @hellkite500. I think its appropriate to make several changes to the design and language used in this feature. First off, its clearly confusing how `key` and `valid_value` are named. Since an `Alias` is intended to be a many to one relationship container, @hellkite500 noted it makes more sense to use `keys` in the place of `valid_value` and `value` in the place of `key`. Hopefully this will make clear the intention of this module for any who come along to use it. Ill edit that in the PR description for congruence.

Second, as @jarq6c pointed out, it would be nice to _test_ if and `Alias` is itself:

```python
x = Alias(1, [2,3])

assert x in x # True
assert 1 in x # True
```

This would also mean the following is valid:

```python
x = Alias(1, [2,3])
y = Alias(1, [3,4])

assert y in x # True
```

Next, @hellkite500 conjured a great idea for semantically "building" `AliasGroup`'s with the `|` or operator. So,

```python
a = Alias(True, "1")
b = Alias(False, "0")

c = a | b # Returns an `AliasGroup`
# or
c = a or b # Returns an `AliasGroup`
```

Personally, I really like this. I lean towards the more idiomatic version with `or`, but either way I really like the semantics! I plan to document this as the preferred way of building `AliasGroup` objects. Just to note it here, in the above example if instead it were `a or a` this would return an `Alias`. Likewise, `a or Alias(True, "2")` is supported and will also return a new `Alias` instance with a `value` of `True` and `keys`, `("1", "2")`. I think this is appropriate given this is non and exclusive or, but im all ears for other opinions.

Lastly, I need to decide on how `AliasGroup` construction is handled internally. Suppose the following:

```python
a = Alias(True, "1")
b = Alias(False, "0")

g = a or b # Returns an `AliasGroup`
# Looks like frozenset(True: frozenset("1"), False: frozenset("0"))

c = Alias(True, "2")

g2 = c or g # Returns an `AliasGroup`
# Should this look like frozenset(True: frozenset("1"), False: frozenset("0"), True: frozenset("2"))
# or frozenset(True: frozenset("1", "2"), False: frozenset("0"))
```

In the latter case, a new `Alias` object is implicitly built for `Alias`'s in the `AliasGroup` with matching `value`s. This would be equivalent to building `a or c or b`. Personally I favor the latter implicit creation as it also covers cases like `a or a or b` which in the first case would be an edge case to check for duplicate `Alias` objects relying on their instance rather than their substance.

aaraney:
This will likely be moved to an issue, but the failure of these tests is due to `nwis_client`'s dependency  `_restclient`. At the moment, `nwis_client` specifies the dependency on `_restclient` as a `git@github` url. I suspect the way the `pip` is resolving this dependency is causing this issue. The issue being, the version of `_restclient` hosted on github is pulled and installed even if the local version of `_restclient` has already been installed.

aaraney:
Been toiling with local docker issues trying to reproduce... I [verified](https://github.com/aaraney/hydrotools/actions/runs/790315804) on a forked branch that switching the dep of `_restclient` from github to pypi in `nwis_client`'s `setup.py` will resolve the issue.

aaraney:
With my previous comment in mind, I am going to squash and merge this feature and open a PR to move `nwis_client`'s dep on `_restclient` over to `pypi`. Ill ref it here after its been merged.

PR #68, opened 2021-04-21
Simplify nwis_client api: `get` method accessable as subpackage level import

aaraney:
This change is up for discussion and I certainly can be persuaded. I think it would be nice to simplify the `nwis_client` api so a user can use the `get` method when importing the subpackage and without having to import `iv`. 

```python
from hydrotools import nwis_client

df = nwis_client.get(
  sites='01646500', 
  startDT='2019-08-01', 
  endDT='2020-08-01'
)
```

This should flatten then overall api. The drawbacks to this are if other nwis rest api functions are desired as future additions, the method name `get` will be reserved for getting nwis data and added functions will have to work around this. Personally if we are to make this change, I think its one we cant really go backwards without really peeving users. However, I do personally prefer a flat api and I think this proposed change could be implemented in other subpackages to flatten out api's for simplicity sake whilst maintaining backwards compatibility.

## Changes

- `nwis_client.IVDataService.get` method now accessible with `nwis_client.get` 

## TODO

- If this feature is desired, update docstrings and readme examples to reflect the change.
- bump minor version

## Checklist

### DIFF
--- a/python/nwis_client/hydrotools/nwis_client/__init__.py
+++ b/python/nwis_client/hydrotools/nwis_client/__init__.py
@@ -1,1 +1,4 @@
 from .iv import IVDataService
+
+# Monkeypatch get classmethod so it's accessable at subpackage level
+get = IVDataService.get


jarq6c:
I understand the motivation, but I'm not sure this is ideal. The Instantaneous Values (IV) web service is not the only REST API offered by USGS/NWIS. See here: https://waterservices.usgs.gov/rest/

All of these services originate at https://waterservices.usgs.gov/nwis/

In the future, the IV service may be the primary product offered by NWIS, but there's no indication they are moving to that soon. They may also support the legacy services for some time.

It would be great to add additional submodule `get` methods for the Site Service or Statistics Service, for example. This change may make that more difficult.

aaraney:
Yeah, this will likely die here. I think your concerns are valid. This change would, as you pointed out, really limit future feature naming. What are your thought on `get_iv` instead? I am not married to the idea and I really dont know if I even like it, but I figured I would at least mention it.

jarq6c:
> Yeah, this will likely die here. I think your concerns are valid. This change would, as you pointed out, really limit future feature naming. What are your thought on `get_iv` instead? I am not married to the idea and I really dont know if I even like it, but I figured I would at least mention it.

I guess it depends on the model we're using to build subpackage import patterns. I assumed an import pattern like `namespace.client.service.method` so for the NWIS Client tool this results in `hydrotools.nwis_client.iv.get` or `hydrotools.nwis_client.statistics.get` for the IV and Statistics services respectively. Also note that for these cases the simplest `get` method represents the most common use case for the module. For `iv` this common use case is retrieving streamflow data in a `pandas.DataFrame` from 1 or more sites.

jarq6c:
Closing this seems non-controversial.

Issue #70, opened 2021-04-27
Implement interface to disable caching for _restclient and nwis_client


jarq6c:
Currently, `nwis_client.iv.IVDataService` sets up caching on initialization. As near as I can tell `_restclient.RestClient` also decides whether to cache at initialization. The result is that there is no native way to disable caching prior to or when requesting data.

Previously, I've successfully used the `requests_cache` context manager to temporarily disable caching, but we might want to offer an interface through the tools themselves.

aaraney:
#86 should have closed this. Closing now

Issue #14, opened 2021-01-13
Disabling requests-cache is not thread safe. Discussion of the future of _restclient's usage of requests-cache 


aaraney:
Due to the way `requests-cache` (used in `_restclient`) is implemented, if the cache has been "installed" and a downstream package uses requests, the downstream package's requests will be cached. Meaning that `requests_cache` implicitly changes the behavior of the `requests` package for all downstream callers. 

Per `requests-cache`'s [documentation](https://requests-cache.readthedocs.io/en/latest/user_guide.html#usage), they do provide a context manager to "disable" the cache. However, as @christophertubbs pointed out [their implementation](https://github.com/reclosedev/requests-cache/blob/6717a336430eb7e6fe17e75c5517f3076ca85b42/requests_cache/core.py#L156) is not thread safe:

```python
 @contextmanager
    def cache_disabled(self):
        """
        Context manager for temporary disabling cache
        ::
            >>> s = CachedSession()
            >>> with s.cache_disabled():
            ...     s.get('http://httpbin.org/ip')
        """
        self._is_cache_disabled = True
        try:
            yield
        finally:
            self._is_cache_disabled = False
```

In discussing this issue with @hellkite500, @hellkite500 mentioned the (actively developed) project [CacheControl](https://github.com/ionrock/cachecontrol). Before active development to fix this behavior, its worth exploring a bit and determining what long term solution to implement.

jarq6c:
With reference to CacheControl, the default cache is a `dict`. They also have a `FileCache` that will persist requests using `pickle` and a `RedisCache`. I'm wondering about performance using the `FileCache` and portability using the `RedisCache`.

aaraney:
> With reference to CacheControl, the default cache is a `dict`. They also have a `FileCache` that will persist requests using `pickle` and a `RedisCache`. I'm wondering about performance using the `FileCache` and portability using the `RedisCache`.


Thanks for doing a little digging... Having not done that so far myself, the issues that you mentioned are problematic and do not sound like they would support our needs. Specifically a `FileCache` does not support our needs nor an in-RAM solution like `Redis`. Additionally, I am not overly keen in bringing in another dependency just to fix this one issue. Plus with the status of `requests-cache`:

> [Noticed last commit was 14-Aug-2019. Which is almost 1.5 years ago as of today.](https://github.com/reclosedev/requests-cache/issues/167#issue-783805523)

[...]

PR #72, opened 2021-04-27
(WIP) Change restclient from monkey patching

aaraney:
In relation to #14 and #70. `_restclient` no longer uses `requests_cache.install_cache` which monkeypatches the `requests` library. Instead, `requests_cache.CachedSession` is now used. `enable_cache` [bool] parameter added to constructor of `RestClient`

## Changes

- `requests_cache.install_cache` is no longer used. Instead `requests_cache.CachedSession` is now used.
- `enable_cache` [bool] parameter added to constructor of `RestClient`

## Notes

- Passes all unittests, however several slow tests for `nwis_client` are now failing. There appears to be something happening with multiprocessing and the `requests_cache.CachedSession` module.

## Checklist

### DIFF
--- a/python/_restclient/hydrotools/_restclient/_restclient.py
+++ b/python/_restclient/hydrotools/_restclient/_restclient.py
@@ -51,8 +51,8 @@ class RestClient:
         if enable_cache:
             try:
                 # Cache requests for 12 hours
-                requests_cache.install_cache(
-                    requests_cache_filename,
+                self._session = requests_cache.CachedSession(
+                    cache_name=requests_cache_filename,
                     backend="sqlite",
                     expire_after=requests_cache_expire_after,
                 )
@@ -61,6 +61,8 @@ class RestClient:
                 error_message = "Something went wrong with setting up `requests_cache`."
                 BaseException(error_message)
                 raise
+        else:
+            self._session = requests.Session()
 
     def get(
         self,
@@ -192,7 +194,7 @@ class RestClient:
             Session
         """
 
-        with requests.Session() as session:
+        with self._session as session:
 
             # retry times and backoff factor (how long to sleep in between
             # retries)
--- a/python/_restclient/hydrotools/_restclient/_restclient.py
+++ b/python/_restclient/hydrotools/_restclient/_restclient.py
@@ -118,7 +118,7 @@ class RestClient:
 
         Examples
         --------
-        >>> 
+        >>>
         """
 
         # Handle default headers.
@@ -139,8 +139,12 @@ class RestClient:
         # Build GET request url
         return requests.Request("GET", url, params=parameters).prepare()
 
-    def Get(self, request: requests.PreparedRequest, **kwargs,) -> requests.Response:
-        """ Thin request.Session.get wrapper. Take prepared request and get
+    def Get(
+        self,
+        request: requests.PreparedRequest,
+        **kwargs,
+    ) -> requests.Response:
+        """Thin request.Session.get wrapper. Take prepared request and get
         response handling errors along the way. Only requests status codes
         200 and 201 returned.If the initial request fails, `self._retries`
         number retries are attempted with a 0.1 backoff factor.
@@ -152,7 +156,7 @@ class RestClient:
         kwargs :
             Keyword arguments passed to requests.Session().send
             See: https://requests.readthedocs.io/en/latest/_modules/requests/sessions/#Session.send
-            
+
 
         Returns
         -------
--- a/python/_restclient/hydrotools/_restclient/_restclient.py
+++ b/python/_restclient/hydrotools/_restclient/_restclient.py
@@ -39,15 +39,16 @@ class RestClient:
         self,
         base_url: Union[str, None] = None,
         headers: Union[dict, None] = None,
-        requests_cache_filename: Union[str, None] = None,
+        requests_cache_filename: Union[str, None] = "cache",
         requests_cache_expire_after: int = 43200,
         retries: int = 3,
+        enable_cache: bool = True,
     ):
         self._base_url = base_url
         self._headers = headers
         self._retires = retries
 
-        if requests_cache_filename:
+        if enable_cache:
             try:
                 # Cache requests for 12 hours
                 requests_cache.install_cache(
--- a/python/_restclient/setup.py
+++ b/python/_restclient/setup.py
@@ -13,7 +13,7 @@ SUBPACKAGE_NAME = "_restclient"
 SUBPACKAGE_SLUG = f"{NAMESPACE_PACKAGE_NAME}.{SUBPACKAGE_NAME}"
 
 # Subpackage version
-VERSION = "2.1.0-alpha.0"
+VERSION = "2.2.0-alpha.0"
 
 # Package author information
 AUTHOR = "Austin Raney"
--- a/python/_restclient/tests/test_restclient.py
+++ b/python/_restclient/tests/test_restclient.py
@@ -109,10 +109,12 @@ def prepared_request_patch(*args, status_code=404, **kwargs):
 
 @pytest.mark.parametrize("status_code", [200, 201])
 def test_get_request(empty_restclient, monkeypatch, status_code):
-    import requests
+    import requests_cache
 
     monkeypatch.setattr(
-        requests.Session, "send", prepared_request_patch(status_code=status_code)
+        requests_cache.CachedSession,
+        "send",
+        prepared_request_patch(status_code=status_code),
     )
 
     assert empty_restclient.Get(None).status_code == status_code
@@ -122,11 +124,14 @@ def test_get_request(empty_restclient, monkeypatch, status_code):
 def test_get_request_exceptions(empty_restclient, monkeypatch, status_code):
     """ Verify that requests.exceptions.ConnectionError is raised """
     import requests
+    import requests_cache
 
     with pytest.raises(requests.exceptions.ConnectionError):
         req = requests.Request("GET", "http://test-url.test/")
 
         monkeypatch.setattr(
-            requests.Session, "send", prepared_request_patch(status_code=status_code)
+            requests_cache.CachedSession,
+            "send",
+            prepared_request_patch(status_code=status_code),
         )
         empty_restclient.Get(req)


aaraney:
Just as a quick note, it appears that the `requests_cache.CachedSession` option is not going to be suitable for usage in `nwis_client` or any other subpackage that uses multiprocessing to retrieve data. I need to do some searching as to why exactly this is the case, but at a high level it is because each session has to acquire a lock to, I would assume, write to the database. I know that python's `sqlite3` library does not support multiprocessing out of the box, so this may be something we will have to address in a cache subpackage as mentioned in #14.

aaraney:
> This is a persistent problem I keep running into and one of the reasons I requested a way to disable the cache. Parallel reads are usually not a problem. Parallel writes are almost always a problem. This is why in `gcp_client` the parallelism is only employed for retrieval and processing. Caching only occurs in the parent process after all the data comes back.

Right (not write). We talked about this offline, but utilizing an async approach or threaded approach will likely be in our best interest. What you described above to implement a caching solution sounds a lot like a "consumer producer" pattern. I think the downfall to that approach is dividing and conquering work after data ingestion. 

I did some testing this morning with `requests_cache.CachedSession` and a `concurrent.future.ThreadPoolExecutor` just to see if I could break it. In the documentation for `requests_cache.CachedSession` it notes that the module is not thread safe. I was not able to break it which I initially _think_ is due to `sqlite3`'s implementation of file locks on the backend that python threads are playing well with. This may not be the case though. I quickly compared the `ThreadPoolExecutor` implementation to using a `multiprocessing.pool` and just a call to `requests.get` and to no surprise, the `ThreadPoolExecutor` was faster (wall clock wise) and had a smaller memory footprint.

jarq6c:
> Right (not write). We talked about this offline, but utilizing an async approach or threaded approach will likely be in our best interest.

I think this will be package specific. For example, `gcp_client` employs a process that retrieves and processes individual NetCDF files. The current package was designed so the retrieval/processing step is effectively CPU bound. We could implement an async alternative, but we would have to reimplement the functionality offered by using the `google_cloud_storage` package and redesign `gcp_client` to separate the retrieval and processing.

Have said that, for any individual package, if async, threading, or multiprocessing can get us from nothing to `DataFrame` faster then we should use that solution (and hope variation in host resources is not a huge factor). This might suggest we need a more robust scaling method to compare these options.

aaraney:
This morning ive done a fair amount of playing around with `multiprocessing`, `threading`, and `async` solutions ranging from simple to complex shared memory implementations. 

Before mentioning things i've tried, Ill add a little context to the problem: The main issue inhibiting `requests_cache.CachedSession` from working with `multiprocessing` is passing a read lock object. Given that on both windows and mac on python 3.8+ the [behavior of the `multiprocessing`](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) library is now to `spawn` instead of (in the mac case) `fork` then `exec`. What this means is a new python instance is started for each worker and is handed a copy of the data needed to complete the task. Meaning a lock object (that cannot be copied) is not passed to jobs as they `spawn` in the processing pool.

Things ive tried:
- Create a new `requests_cache.CachedSession` `spawn`ing a new job. This works, however the overhead needed to do this slows down things when both getting and caching or just retrieving a cached object. I think this could be a work around for some that don't want to implement a more complex solution (ill get to that later). Im not sure that I want to implement this behavior in `_restclient` as the default as it will hurt performance for all users, not just those using multiprocessing.
- Using a `concurrent.futures.ThreadPoolExecutor` to share a `requests_cache.CachedSession` session, but perform `get` calls in their own thread of execution. This works and for the most part and in all cases has a lower memory footprint than any `multiprocessing` implementation (not surprising), however through repeated `get` calls (e.g. `pool.map(session.get, [<1>, <2>, <3>]`) I found that occasionally it would take upwards of 20 seconds wall time to retrieve the data. I verified that I was not being throttled, so I am not exactly sure why this is occurring. I assume this is due to a race condition involving the connection to the `db`.
- Next, I tried using `asyncio` by wrapping the `requests_cache.CachedSession.get` call in a `loop.run_in_executor` call. Having looked in to the implementation of `loop.run_in_executor`, this is nearly identical to the `ThreadPoolExecutor` example I mentioned prior. When running in an executor, the work just gets pushed to a `ThreadPoolExecutor`. As such, I received the same behavior I did above. Sporatic and long execution times that I assume involve writing/reading to/from the `db`.
- Lastly, I looked outside of `requests_cache` for a purely `async` solution. This led me to [`aiohttp`](https://docs.aiohttp.org/en/stable/) and [`aiohttp-client-cache`[(https://github.com/JWCook/aiohttp-client-cache). I recognized a few of the contributors to `aiohttp-client-cache` from `requests_cache`, so that is a good sign. In short, `aiohttp-client-cache` is to `aiohttp` as `requests_cache` is to `requests`. I setup the same tests i'd been running in the above listings and found that that `aiohttp-client-cache` is typically faster than `multiprocessing` and has a much lower memory footprint. The library supports the `redis`, `mongo`, `sqlite`, and `dynamo` `db`s. Personally, at this moment in time, I think switching to both `aiohttp` and `aiohttp-client-cache` make sense. So what does that mean? Well it would mean we would want to support `async` functionality and have wrappers to `async` functions that act like serial functions. Meaning you could use `_restclient` in both cases. I think this kind of refactoring is a priority, but ill yield to others input first.

Moving forward, I think it is in our best interest to treat i/o bound work differently from cpu bound work. This means, using `async` and `threads` for i/o bound work and `multiprocessing` when we need it. If for example, data is needed from the web then it is processed (i.e. `nwis_client`) it makes more sense to consume data from the web using a `async` approach and then hand off processing to either the main thread or to a process pool (consumer producer semantics). I think we can document this pattern well in some examples along the way to serve as a starting point for ourselves and others trying to accomplish a similar goal.

jarq6c:
> Moving forward, I think it is in our best interest to treat i/o bound work differently from cpu bound work. This means, using async and threads for i/o bound work and multiprocessing when we need it. If for example, data is needed from the web then it is processed (i.e. nwis_client) it makes more sense to consume data from the web using a async approach and then hand off processing to either the main thread or to a process pool (consumer producer semantics). I think we can document this pattern well in some examples along the way to serve as a starting point for ourselves and others trying to accomplish a similar goal.

In general, I agree.

However, I think this only makes sense if the data retrieved fits in memory. If it does not and we have to write the data to disk for later processing are we still realizing the benefits offered by asyncio? Many of the asynchronous implementations I've seen are handling lots of relatively small payloads. `nwis_client` and `gcp_client` may be handling lots of requests for relatively large amounts of data. How do we balance rapid retrieval and limited memory?

The current paradigm implicitly assumes the data are retrieved and quickly pared down and optimized (the CPU bound part). Where do all the data live while waiting to be optimized?

aaraney:
> However, I think this only makes sense if the data retrieved fits in memory. If it does not and we have to write the data to disk for later processing are we still realizing the benefits offered by asyncio? Many of the asynchronous implementations I've seen are handling lots of relatively small payloads. nwis_client and gcp_client may be handling lots of requests for relatively large amounts of data. How do we balance rapid retrieval and limited memory?

That's a fair point and I think it depends on _how_ that data that cannot fit in memory is stored. If we use multiprocessing, I think things will become more complicated if we deal with "caching" data to a centralized store (a file db, h5 store etc.). Trading lock objects through shared memory I think will likely get messy and overall bog things down. I could also be very wrong (@hellkite500 / @christophertubbs can you weigh in?). In the case of `async`, my intuition says it will actually be easier to handle saving to disc along the way if required. A new thread pool executor can be spun up and handle the work while other jobs are still underway. Likewise, `async` reads may be advantageous of any data that is stored locally, that data can be asynchronously consumed and placed onto a process pool executor for cpu bound work. As for the scalability of `async`, which is what im taking the latter part of your comment to mean, in tests ive run comparing `aiohttp` to `multiprocessing` with `requests`, `aiohttp` pretty much always wins and is never drastically slower even as calls grow in size. I just performed an experiment looking at retrieving 10 years of nwis data from 10 sites using both approaches and `aiohttp` with an enabled cache, still was 5 to 7 seconds faster than `requests`.  This is still a "small" test, but still its somewhat meaningful.

jarq6c:
From our previous meeting, it seems apparent this is the way to go. As a benchmark would you like to test the following request against your async method?
```python
from hydrotools.gcp_client import gcp
from hydrotools.nwis_client.iv import IVDataService
from functools import partial
import timeit

model_service = gcp.NWMDataService()
sites = model_service.crosswalk['usgs_site_code']

data_request = partial(
    IVDataService.get,
    sites=sites,
    startDT='2021-01-01T01:00Z',
    endDT='2021-01-01T18:00Z'
)

t = timeit.timeit(data_request, number=1)
print(t)

# System: 6 cores/6 threads
# First run: 83.5404682229273
# Second run: 13.354135259054601
```

jarq6c:
Additional result:

Python 3.9
System: 24 cores/24 threads
First run: 120.14240689598955
Second run: 19.547235067002475

aaraney:
I just ran the same test, however just looking at the io time. So this does not include serializing the data into a dataframe.

```python
#!/usr/bin/env python3

from hydrotools.gcp_client import gcp
from hydrotools.nwis_client.iv import IVDataService


async def get_json(session, url):
    async with session.get(url) as resp:
        return await resp.json()


async def mget(urls):
    headers = {"Accept-Encoding": "gzip, compress"}
    async with ClientSession(headers=headers) as session:
        work = [get_json(session, url) for url in urls]
        return await asyncio.gather(*work)


if __name__ == "__main__":
    from functools import partial
    import asyncio
    from aiohttp import ClientSession
    import timeit

    model_service = gcp.NWMDataService()
    sites = model_service.crosswalk["usgs_site_code"].to_list()
    url = "https://waterservices.usgs.gov/nwis/iv/?format=json&sites={}&startDT=2021-01-01T01:00%2b0000&endDT=2021-01-01T18:00%2b0000&parameterCd=00060&siteStatus=all"

    sites = [url.format(",".join(sites[s : s + 50])) for s in range(0, len(sites), 50)]

    loop = asyncio.get_event_loop()
    work = partial(loop.run_until_complete, mget(sites))

    t = timeit.timeit(work, number=1)
    print(t)
```

System: 4 cores / 4 threads
Download time: 10.84 seconds

aaraney:
In the context of the above findings and the previous discussion, I think this PR is going to die here. I am currently working on extending `aiohttp` with a cache client and an exponential backoff. I plan to refactor the `RestClient` to utilize that class and squash this feature in the process. The above conversation is worthy of pointing back to in the future.

PR #73, opened 2021-05-01
Add Simple HDF5 Backed Caching Utility

jarq6c:
Adds a simple `@hdf_cache` decorator for use by methods that return `pandas.DataFrame`. The HDF5 file and dataset keys are set by the user when the method is called. The decorator wraps the original method and looks for special arguments (`hdf_cache_path` and `hdf_cache_key`) used to save the dataframe to a file. If a wrapped method is called without these arguments a `UserWarning` is raised.

I consider this the first of many possible caching `utilities` we could implement for `DataFrame` or other objects. This PR is mostly the result of growing personal resentment I had rewriting this code for the nth time in my own workflows. This design pattern is suited to data science workflows where the results of long running processes are saved to interim files for later retrieval during development and data exploration.

## Additions

- `hydrotools.utilities` subpackage
- `hydrotools.utilities.hdf_cache` module containing the `@hdf_cache` decorator function
- `hydrotools.utilities.stock_dataframes` containing a `random_dataframe` method that demonstrates use of the `@hdf_cache` decorator

## Removals

- None

## Changes

- None

## Testing

1. `test_stock_dataframes` and `test_hdf_cache` test various permutations of cached and non-cached calls, as well as warnings that are raised.

## Notes

- None

## Todos

- None. Perhaps develop criteria to determine what makes a good candidate "utility"

## Checklist


aaraney:
I really like the idea of an HDF5 cache, however I think we can make some improvements to the semantics of the decorator and also a context manager. As it sits now, my main concern with the decorator implementation is the loss of type hints post decoration. Meaning that the `hdf_cache_path` and `hdf_cache_key` are implicit kwargs that the user has no clue exists by way of type hints. 

```python
import inspect

from hydrotools.utilites.stock_dataframes import random_dataframe

print(inspect.signature(random_dataframe) 
# returns,
(start: Union[pd.Timestamp, str] = '1970-01-01T01',
rows: int = 10,
freq: Union[str, pd.tseries.offsets.DateOffset] = '1H',
float_column_name: Union[str, List[str]] = ['value'],
datetime_column_name: str = "value_time")
```

As such, the `hdf_cache_path` and `hdf_cache_key` kwargs are implicit to the function signature.

Ive been using the [`forge`](https://python-forge.readthedocs.io/en/latest/)  to get around a similar problem in my fork of `_restclient`. In short, the `forge` library lets you forge the signature of a method (adding args from another method, remove an arg, binding a default arg value to a method, etc.). 

For example:

```python
from functools import wraps
from typing import Callable
import forge

def append_signature(target: Callable, *, include=None, exclude=None):
    def dec(func: Callable):
        @wraps(func)
        @forge.compose(
            forge.copy(target, include=include, exclude=exclude), # copy target function signature
            forge.insert(forge.fsignature(func), index=0), # prepend signature of func to target function signature
        )
        def inner(*args, **kwargs):
            # will need to adjust things here to get `target` to work.
            # Its probably best to make an aggregate decorator that appends the signature
            # the applies the functionality of `target` to `func` all in one aggregate decorator
            return func(*args, **kwargs)

        return inner

    return dec

def cache(func: Callable, *, key=None, path=None):
    r = func()

    if key is not None and path is not None:
      # do cache stuff
      pass

    return r

@append_signature(cache, exclude="func")
def work(x: int, y: int):
  return x + y

import inspect
print(inspect.signature(work))
# returns,
(x: int, y: int, *, key=None, path=None)
```

In this form, type hints are maintained and `append_signature` is generalized so it could be used as its own utility. Personally, I like the idea of either including generalized decorators in this `utilities` subpackage or breaking them out into a different subpackage. The only thing missing here is creating a dynamic docstring, meaning `help(work)` still would not include any information about the `key` and `path` args, if `cache` were to have a docstring. That sounds like another decorator to me, but it again, sounds like a good candidate for this or another subpackage. 

With all of the above being said, I think it would be nice to capture caching functionality in a context manager. I am thinking something like:

```python
with hdf5cache(key="key", path="/my/path/store.h5") as cache:
  # some intensive work
  df = tae_bo()
  cache.write(df)
```

In the above example, a context manager would define a transaction to the hdf5 store although it is limited a single transaction. I think we could mutually support the following:

```python
with hdf5cache() as cache:
  for task in jobs:
    # some intensive work
    df = tae_bo(task)
    # write could be treated as a generator which work is yielded to
    cache.write(df, key="key", path="/my/path/store.h5") 
```

We should be able to capture all of the above use cases (as decorator and context manager) through as class that implements the `__call__`, `__enter__`, and `__exit__` dunder methods. What are your thoughts?

python/utilities/tests/test_hdf_cache.py
```diff
@@ -0,0 +1,43 @@
+import pytest
+from  hydrotools.utilities.hdf_cache import hdf_cache
+import pandas as pd
+import numpy as np
+from pathlib import Path
+import warnings
+
+def test_hdf_cache():
```
aaraney, line 8: We may want to add some tests that write to a an hdf store and read back from it just to verify that things are writing correctly.

aaraney (changes requested):

jarq6c:
Hmm, maybe "cache" was the wrong term to use for what this is supposed to do. The use-case I was considering was running _and_ re-running long computational or data processing workflows. The idea of this feature is that you have a function that produces a DataFrame from a long process. You key that long process to a dataframe in an HDF file, so that every time you call that function it only runs the long process if the DataFrame is not in the HDF file. This could be easily implemented just by surrounding every function call with the code inside the decorator function, but I was trying to implement some syntactical sugar to avoid having to rewrite that code everywhere.

I'll mull over the design pattern a bit more in the context of your recommendations.

aaraney:
> Hmm, maybe "cache" was the wrong term to use for what this is supposed to do. The use-case I was considering was running _and_ re-running long computational or data processing workflows. The idea of this feature is that you have a function that produces a DataFrame from a long process. You key that long process to a dataframe in an HDF file, so that every time you call that function it only runs the long process if the DataFrame is not in the HDF file. This could be easily implemented just by surrounding every function call with the code inside the decorator function, but I was trying to implement some syntactical sugar to avoid having to rewrite that code everywhere.
> 
> I'll mull over the design pattern a bit more in the context of your recommendations.

Ah I see, no that would also be a cache, I guess technically a Memoization. I think calling it a cache is acceptable. I am thinking in the context of a cache that is time enabled, meaning after some time you recache the function call. What you described is what we need (IMO).

jarq6c:
After extensive conversation with @aaraney I've decided to reimplement this a context manager.

PR #77, opened 2021-05-11
Add Caches Subpackage with Minimal HDF5 backed caching utility

jarq6c:
Take 2! This a slightly scaled back version of an HDF5-backed caching class implemented as a context manager. This class is a fairly thin wrapper around `pandas.HDFStore` that implements a `HDFCache.get` method to run a long process and store the results in a HDF5 file for later retrieval if `HDFCache.get` is called again with the same key.  The original functionality of `HDFStore.get` and other `HDFStore` methods are still accessible via the `HDFCache.store` property. Handling key logic and strictly associating keys with certain functions is left to the user. This allows users to build their own system of keys to suit their data.

## Additions

- `HDFCache` class
- `hydrotools.caches` subpackage where other caches can live @hellkite500 

## Removals

- None

## Changes

- New subpackage

## Testing

1. Tests the caching by storing a dataframe with random values and verifying equality.

## Notes

- The internal `HDFStore` used by `HDFCache` default to `format='table'`. This is because `pandas.Categorical` is incompatible with `'fixed'` format and HydroTools uses categoricals everywhere.

## Todos
- May want to implement more advanced cache management later. For now, a lot of functionality can be implemented using the internal `pandas.HDFStore` and `pytables`.

## Checklist


jarq6c:
Note one oddity: the `pandas.HDFStore` docstring gives `str` as the type of the `path` parameter. However, it still works with `pathlib.Path`. I also gave `str` as the corresponding type hint in `HDFCache` for consistency.

python/caches/hydrotools/caches/hdf.py
```diff
@@ -0,0 +1,108 @@
...
+Classes
+---------
+ - HDFCache
+
+"""
+
+import pandas as pd
+from typing import Callable
+
+class HDFCache:
```
aaraney, line 17: This looks great, just curious why the `*args` `**kwargs`? It seems more appropriate to include all the args [`HDFStore`](https://github.com/pandas-dev/pandas/blob/ea05559dc748eafe7de4f3ea39401b2d3f21bb4e/pandas/io/pytables.py#L556) includes.

aaraney: Additionally, it would be nice to include the `__del__` method for the garbage collector in the case of using `HDFCache` without a context manager.

python/caches/tests/test_hdf_cache.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+from hydrotools.caches.hdf import HDFCache
+import pandas as pd
+import numpy as np
+from pathlib import Path
+
+cache_path = Path('test_cache.h5')
```
aaraney, line 7: For testing purposes, it might be best if we move over to using `NamedTemporaryFile`s. It would probably be best to either return the name of a temporary file from a fixture or do it within the context of a test itself.

Other tests that would be nice to see:
- using a byte string as a key.
- using different string encodings as keys of the same name
- Getting a `df` from the cache outside of the context of a context manager (from an instance of `HDFCache`).

aaraney (approved):
In general, I think this is a great addition. I like the context manager implementation over #73. I am going to tentatively approve this PR given the small optional changes I mention in my comments. I do think it would be worth our time to add a few more unittests to cover some "edge" cases.

python/caches/hydrotools/caches/hdf.py
jarq6c: My approach to `*args` and `**kwargs` was to limit explicit parameters to those required (`path`) and those where I change the default (`format`). The reason is allow flexibility if more parameters get added to `pandas.HDFStore` later. Additionally, `HDFStore` itself passes arguments on to pytables, so I figured was easier to limit explicit parameters to those essential to the functioning of `HDFCache`.

Roger that on `__del__`.

aaraney: Yeah and that's a valid point. Maybe we can lean on the [`forge`](https://github.com/dfee/forge) library to automagically bring over those type hints. Its yet another dep to bring in, but im using it over in `_restclient` and it really small, so I am not overly concerned about bring it in to the project. Thoughts? To your point, I hope that future versions of python include decorators or some way to extend/share type hints more easily. `inspect` is not cutting it at the moment.

Cool cool 👍

python/caches/tests/test_hdf_cache.py
jarq6c: I used a `TemporaryDirectory`. `tempfile` returns file buffers and `HDFStore` needs a path string.

I added a couple more tests and now raise for non-string keys.

python/caches/hydrotools/caches/hdf.py
jarq6c: It seems like the issue with wrapping functions is a contentious one without a one-size-fits-all solution.

https://hynek.me/articles/decorators/

aaraney: Yeah, decorators are just so flexible that I would assume conversations quickly dissolve into users returning to their opinion corner instead of actually coming to a solution of sorts. Nice article though, it seems that `functools` has addressed many of the issues raised in that article -- which is good. However, using decorators to extend function args almost like you would in an inherited class or mixin still has not been agreed upon enough to make its way into the standard library.

python/caches/tests/test_hdf_cache.py
aaraney: Okay great! Thanks for doing that.

aaraney:
Rebase and merge at will

PR #82, opened 2021-05-20
Update caches with setup metadata and required files

jarq6c:
Pushing through without review. Updates `setup.py` metadata and adds OWP required files at the subpackage level.


jarq6c:
Passed all tests.

Issue #65, opened 2021-04-12
Document Canonical Format


jarq6c:
As I'm revisiting the `nwis_client` and `gcp_client` tools, I think this is a good time to be daring and document the fixed vocabulary that describes `hydrotools` canonical dataframes. I'm open to discussion on this topic, but so far I'm leaning toward the column definitions given below. These definitions are not exactly compatible with our internal services, but our internal services are not exactly compatible with each other. My motivation was to establish a vocabulary for `hydrotools` that was consistent and sufficiently descriptive. Not all of these columns will be present or relevant to all dataframes, but where present definitions should be consistent.

There are two possible breaking changes, `value_date` is now `valid_time` and `start_date` is now `reference_time`.

These definitions also lean a lot on categorical types to avoid the use of multi-indexes. However, nothing prevents users from recasting these dataframes to use multi-indexes or from adding new columns to their data (like a `custom_site_identifier` column for example). These column labels are just meant to cover and define what the various client tools might return.

@aaraney @hellkite500 

### HydroTools Canonical DataFrame Column Definitions
`value` [*float32*]: Indicates the real value of an individual measurement or simulated quantity.
`valid_time` [*datetime64[ns]*]: formerly `value_date`, this indicates the valid time of `value`.
`variable_name` [*category*]:  string category that indicates the real-world type of `value` (e.g. streamflow, gage height, temperature).
`usgs_site_code` [*category*]: string category indicating the USGS Site Code/gage ID
`nwm_feature_id` [*category*]: string category indicating the NWM reach feature ID/ComID
`nws_lid` [*category*]: string category indicating the NWS Location ID/gage ID
`usace_gage_id` [*category*]: string category indicating the USACE gage ID
`measurement_unit` [*category*]: string category indicating the measurement unit (SI or standard) of `value`
`qualifiers` [*category*]: string category that indicates any special qualifying codes or messages that apply to `value`
`series` [*integer32*]: Use to disambiguate multiple coincident time series returned by a data source.
`configuration` [*category*]: string category used as a label for a particular model simulation configuration (e.g. short_range, medium_range)
`reference_time` [*datetime64[ns]*]: formerly, `start_date`, some reference time for a particular model simulation. Could be considered an issue time, start time, end time, or other meaningful reference time. Interpretation is simulation or forecast specific.
`longitude` [*category*]: float32 category, WGS84 decimal longitude
`latitude` [*category*]: float32 category, WGS84 decimal latitude
[...]

PR #83, opened 2021-05-20
Update default caching functionality in `gcp_client`

jarq6c:
This PR adds incremental improvements to how `gcp_client` handles the local cache of NWM model data. The default caching functionality now uses `hydrotools.caches.hdf.HDFCache`. The ability to disable caching at the `get` level has been added, as has been the ability to specify the path for the cache and  a default root HDF5 group to use.

## Additions

- New `cache_path` and `cache_group` options set when the `NWMDataService` object is instantiated. Defaults to `gcp_client.h5` and `gcp_client`, respectively.
- New `cache_data` keyword argument for `get` to control whether the retrieved data is cached or not. Default behavior is to cache.

## Removals

- Old caching code.

## Changes

- Switched to `HDFCache` for default caching
- Old `get` method is now called `get_cyle`. The new `get` method wraps the old with a conditional that determines whether to cache or not
- `valid_time` is now `value_time` consistent with discussion in #65 

## Testing

1. Test `cache_path` and `cache_group` getters and setters
2. Test data retrieval using independent `pandas.HDFStore` (key pattern is documented)
3. Test disable cache

## Notes

- Seems a reasonable approach for various methods to lean on a default `cache`, so long as the ability to disable and/or customize the default behavior is implemented. This allows users to either use the tool without need to worry about caching, use a centralized cache, or implement their own solution.

## Todos

- Still need Alaska crosswalk for NWM v3.0

## Checklist


python/gcp_client/CONTRIBUTING.md
```diff
@@ -0,0 +1,176 @@
+# Guidance on how to contribute
```
aaraney, line 1: We may want to try using `git`'s symlink tracking for these files. I can see down the road versions of this document changing. Just thinking for long term maintainability/laziness.

python/gcp_client/LICENSE
```diff
@@ -0,0 +1,7 @@
+“Software code created by U.S. Government employees is not subject to copyright
```
aaraney, line 1: LICENSE can also probably be symlinked. It will be worth testing how `setuptools` plays with the symlink when you for example build a wheel.

python/gcp_client/hydrotools/gcp_client/gcp.py
```diff
@@ -46,7 +49,9 @@ def __init__(
         self, 
         bucket_name: str = 'national-water-model', 
         max_processes: int = None,
-        location_metadata_mapping: pd.DataFrame = None
+        location_metadata_mapping: pd.DataFrame = None,
```
aaraney, line 52: To preface this, I am also beginning to change the way I handle keyword arguments... With that being said, it may be in our best interest and the interest of any users that may start using this, to make default keyword arguments keyword only args. Im sure you are aware that in python these although are visually "keyword" arguments, python will still implicitly treat them as positional arguments if you don't provide the variable name in the call. So, to make it an explicit keyword arg you have to do the following:

```python
def function(positional_only, /, positional_or_keyword="arg1", *, keyword_only="arg2"):
  ...
```

aaraney: I think we should start adopting this way of writing and maintaining method signatures so this kind of change would not be a breaking change to the API. Thoughts?

python/gcp_client/setup.py
```diff
@@ -14,7 +14,7 @@
 SUBPACKAGE_SLUG = f"{NAMESPACE_PACKAGE_NAME}.{SUBPACKAGE_NAME}"
 
 # Subpackage version
-VERSION = "2.3.4"
+VERSION = "2.4.0"
```
aaraney, line 17: Given the above changes in canonical dataframe output and a few minor changes to the api, I think this really "should be" a major release bump. But if you disagree, id love to hear your thoughts!

aaraney (approved):
Glad to see relevant non-source files migrated to this subpackage. I left a comment regarding how we may want to migrate this files in a comment, so I will leave the conversation there.

In general this looks great, I am going to approve it, but would like to talk through my comments before the merge.

python/gcp_client/hydrotools/gcp_client/gcp.py
jarq6c: Some references:
https://docs.python.org/3/whatsnew/3.8.html#positional-only-parameters
https://www.python.org/dev/peps/pep-0570/

Seems like a design choice. This is a case where I could see this being the signature:
```python
def __init__(bucket_name="national_water_model", /, max_processes=None, *, location_metadata_mapping, ...):
```
But where exactly to place `/` and `*` is up for debate. The ability to fix the API with positional-only parameters is a boon, but there would still need to be thought put in to how you want the API to actually function. It's nice to have this option, but I'm not sure if it represents something we would apply broadly (at least not yet). I'd like to experiment more with these options to get a feel for the design patterns and uses that would benefit most (as well as those that might not).

Adopting this would lock us into Python 3.8+ right?

python/gcp_client/setup.py
jarq6c: I think you're right

python/gcp_client/CONTRIBUTING.md
jarq6c: Done. Seems to play fine with setuptools

python/gcp_client/hydrotools/gcp_client/gcp.py
aaraney: I should have excluded the positional only argument from my example (`/`), I think my inclusion probably added confusion. I was just thinking about the keyword only arguments vs keyword/positional arguments (see [PEP 3102](https://www.python.org/dev/peps/pep-3102/)). I think adopting this pattern would lead to less breaking changes down the road. So, you above example would be:


```python
def __init__(bucket_name="national_water_model", max_processes=None, *, location_metadata_mapping, ...):
```

The positional only feature is 3.8+ but keyword only was introduced in 3.0.

jarq6c: gotcha. implemented

Issue #50, opened 2021-03-16
nwis_client: Retrieving large amounts of data takes too long


jarq6c:
Case study: A user needs to conduct a long evaluation and wants 25 years of streamflow data from 2000+ USGS gage locations.

Obstacles:

1. We currently only parallelize by site, can we parallelize by chunking in time?
2. How do we get around limited memory?
3. Will this break `requests_cache`?
4. Can a `DataFrame` cache help us here?

Based on a use-case from @hellkite500 

@aaraney not necessarily looking for solutions here yet. Just wanted to start the discussion. The `nwis_client` tool was really the first tool we fleshed out. So, it seems fitting to start discussions about scaling `evaluation_tools` here.

aaraney:
I have conflicting views about this. On one hand, I totally get the desire to have the Swiss army knife of data retrieval tools, but on the other hand I think expecting that one software can fulfill every use case without having to write a bit of code yourself is just not realistic. 

> @aaraney not necessarily looking for solutions here yet.

Well below are some unsolicited starting solutions.

> Obstacles:
> 
> We currently only parallelize by site, can we parallelize by chunking in time?

I think we can achieve this by using a defined time period. So for example, time periods greater than 1 year are chunked by year. I don't see this being an option that the user can specify like the `partition_max_size` we have now that determines when to split by site. I make this point because dealing with time just gets messy thanks in part to the Romans. Chunking by month also seems too granular, but I'm happy to hear what others think about this and other proposed approaches.

> How do we get around limited memory?

This is a more difficult question to answer. I guess first off, to answer this you've got to come up with a system agnostic heuristic that answers, "should I save too disk or not?" Additionally, when does the software ask this question and how often? Or is it just a one time thing (at run time)? I just think there are a lot of edge cases that require speaking to if we want to fully satisfy this question.

As for how can we actually get around the memory barrier, I think a simple sqlite db should suffice. As workers complete a job instead of leaving their response in memory, they write to a local sqlite db that can be queried. The sqlite db could reflect the `evaluation_tools` canonical dataframe. I think there is a valid argument for using `HDF` here or something like `feather` and im open to hear those arguments, just trying to stay true to the `evaluation_tools` mission of tools for data scientists. SQL syntax, while not as common in hydrology in my experience, is a tool most data scientists have.

> Will this break requests_cache?

If you mean once you make the call that hits `NWIS` if you make the same identical call will it hit `NWIS` again? That should still work properly.

[...]

Issue #79, opened 2021-05-12
As a developer, I would like to specify development dependencies at the subpackage level


aaraney:
At the moment, all development deps are installed at the namespace package level, namely `pytest`. I would like to come to an agreed upon standard for specifying development deps to `setuptools`. Below are two examples `setup.py` file implementations that solve this problem:

1.
```python
# file: setup.py

from setuptools import setup
from setuptools.command.develop import develop
import subprocess

DEVELOPMENT_REQUIREMENTS = ["pytest"]

# Development installation
class Develop(develop):
    def run(self):
        # Install development requirements
        for dev_requirement in DEVELOPMENT_REQUIREMENTS:
            subprocess.check_call(
                [sys.executable, "-m", "pip", "install", dev_requirement]
            )
        develop.run(self)

setup(
    name="mypackage",
    description="some package that does something",
    install_requires=["pandas"],
    cmdclass={"develop": Develop},
)
```

The above is used as such, `python setup.py develop`. This will install all the deps, development deps, and the package in an "editable" form (i.e. like `pip install -e  .`).

2. 
```python
# file: setup.py

from setuptools import setup

DEVELOPMENT_REQUIREMENTS = ["pytest"]

setup(
    name="mypackage",
    description="some package that does something",
    install_requires=["pandas"],
    extras_require={"develop" : DEVELOPMENT_REQUIREMENTS},
)
```

2 is used as follows, `pip install -e ".[develop]"` (the quotes are not required in bash, but are required in zsh and likely other shells).

Personally, I prefer the second option over the first. The second option uses `pip` and thus files like `pyproject.toml` are regarded, whereas in 1, they are not. Likewise, this functionality should work on `pypi`, I am unsure at this moment how you would specify to include tests in the package though, so that may be a moot point. The second solution is also far less code and more maintainable across subpackages IMO.

aaraney:
@jarq6c can you weigh in on this with your opinion?

jarq6c:
Based on a cursory exploration of the current Python packaging landscape, I'm also partial to option 2. This will move us in the direction of explicitly leaning on `pip` to handle packaging for any purpose (development, release, etc).

aaraney:
I agree moving under the umbrella of `pip` should only help us as `hydrotools` grow. I will open a PR and ref it here moving all subpackages under option 2 in due time. Ill leave this open to track the progress though and document any snags that occur.

PR #86, opened 2021-06-04
Nwis client refactor / restclient urllib quote_treatment, safe, and quote_overide_map / restclient mget supply parameters/headers as collection feature additions

### LINKED ISSUE #70: Implement interface to disable caching for _restclient and nwis_client

aaraney:
Refactor NWIS Client to use latest version of `_restclient`. Admittedly, this PR is not limited in scope, I got carried away updating code in `hydrotools._restclient` to support my needs in the nwis refactor. As such the below change log has been split to distinguish between the two scopes.

The nwis client implementation is now more modular and general. I implemented a factory like pattern that should be easy to extend and maintain in the future. I may refactor the factory to use a enum in the future, but for now this is a good working implementation. 

nwis client now uses asyncio on the backend via `_restclient` and as such, all multiprocessing code has now been removed.

The client now supports getting multiple of any "major filter" (sites, states, hucs, counties, bounding boxes) -- meaning you can start to ask for a whole lot of data more easily. All major filters support using strings and lists as inputs or common scientific data structures (`np.array`, `pd.Series`). 

closes #70
makes #50 pretty easy to implement 

### Additions

-`IVDataService` parameters: `enable_cache` and `cache_expire_after` allowing control over cache usage.

### Removals

- `IVDataService` parameters: `process` and `retry` removed.
- All code using multiprocessing removed
- `IVDataService` method following pattern `get_raw_` and `get_json` removed
- U.S. State constants removed and no longer checked
- `procs` property removed

### Changes

- `IVDataService` now uses aiohttp via `_restclient` to do asynchronous retrieval of data. 
- documentation updated with examples
- `IVDataService::get` 
  - now accepts multiple states, hucs, counties, bounding boxes
  - is now an instance method instead of a `classmethod`
- The default splitting factor for `sites` is now 20 instead of the previous 100.

### Additions

- `Url` keyword arguments `quote_treatment`, `safe`, and `quote_overide_map` added to support flexibility in quoted url representation. Urls are still comparable using the `==` operator however urls can differ in their `quote_url` representation.
- `RestClient.mget` now accepts a collection of query parameters and/or headers. It is not required to provide urls to `mget` if base url is set. Thus, a collection of query's can be passed in a collection and the will be built into requests as parameters of the base url. 

### Removals

-

### Changes

- Update `Url` docstring to reflect new kwarg additions

## Testing

1. The overwhelming  majority of added features are tested in this PR. Previous unit tests have been modified to test changed implimentation.

## Notes

-

## Todos

- [ ] ~regen sphinx docs~
- [x] ~Verify~ that major filters undergoing splitting are sorted for caching purposes.

## Checklist

- [ ] Changes are limited to a single goal (no scope creep)

Edit: 
- fix typos
- add `_restclient` related changes to log that didn't make the first cut.


python/_restclient/tests/test_restclient.py
```diff
@@ -141,3 +141,12 @@ def test_get_headers_have_precedent_over_instance(basic_test_server):
```
jarq6c, line 141: Both `test_restclient.py` and `test_client.py` are failing on my machine with: "fixture 'aiohttp_raw_server' not found"

jarq6c: Similar issue with `test_nwis.py`

python/nwis_client/hydrotools/nwis_client/iv.py
```diff
@@ -12,58 +12,112 @@
...
+    sqlite3 request cache and asynchronous request backend that splits up and retrieves
+    requests concurrently.
 
     Parameters
     ----------
-    processes : int
-        Max multiprocessing processes, default 3
-    retry : int
-        Max number of, per request retries, default 3
+    enable_cache : bool
```
jarq6c, line 37: Setting this to `enable_cache=False` does not appear to do anything. `nwis_cache` was still created.

jarq6c (changes requested):
This is an awesome amount of work and an important improvement in functionality. I am thoroughly impressed! I found some issues, but I'm not sure if it's because I don't know what I'm doing or they are real bugs.

I was unable to run the tests locally. `pytest` had some issues with how the parameterized fixtures are setup. This was in Python 3.8.5. I was able to get the code installed and run, but not tested. 

I ran the code below in an attempt to get data without using a cache, but the cache was still created.
```python
# Import the NWIS IV Client
from hydrotools.nwis_client.iv import IVDataService

# Instantiate service
obs_service = IVDataService(enable_cache=False)

# Retrieve data from a single site
observations_data = obs_service.get(
    sites='01646500', 
    startDT='2019-08-01', 
    endDT='2020-08-01'
    )

# Look at the data
print(observations_data.head())
```

aaraney:
> This is an awesome amount of work and an important improvement in functionality. I am thoroughly impressed! I found some issues, but I'm not sure if it's because I don't know what I'm doing or they are real bugs.


Thanks! Im pretty pleased with how it has turned out so far. Having looked into the things that you found below, rest assured I made the problems haha.

> I was unable to run the tests locally. `pytest` had some issues with how the parameterized fixtures are setup. This was in Python 3.8.5. I was able to get the code installed and run, but not tested.

Yeah this is related to #79. I need to include the `pytest-aiohttp` as a development dep for the package. Note that to install the package in "development mode" you will now do `pip install -e "python/nwis_client[develop]"`.

> I ran the code below in an attempt to get data without using a cache, but the cache was still created.
> 
> ```python
> # Import the NWIS IV Client
> from hydrotools.nwis_client.iv import IVDataService
> 
> # Instantiate service
> obs_service = IVDataService(enable_cache=False)
> 
> # Retrieve data from a single site
> observations_data = obs_service.get(
>     sites='01646500', 
>     startDT='2019-08-01', 
>     endDT='2020-08-01'
>     )
> 
> # Look at the data
> print(observations_data.head())
> ```

Great catch, the culprite seems to be `IVDataService::get`'s implementation as a class method. `get` is creating its own instance of the class which by default will cache things. We have two options to get around this:

1. Make `IVDataService::get` a regular instance method. Meaning you would have to use it the way shown above.
2. or, add the arguments in the constructor to `IVDataService::get`. So, you would have to pass `enable_cache=False` too `IVDataService::get` to disable the cache. 

Im not really satisfied with either option, but I think the first is probably the better of the two IMO. Thoughts?

aaraney:
What would be great is if `IVDataService::get` had `@classmethod` and instance method versions that were callable in either situation, but i'm not sure exactly how to accomplish that. Im thinking about it though.

jarq6c:
> Great catch, the culprite seems to be `IVDataService::get`'s implementation as a class method. `get` is creating its own instance of the class which by default will cache things. We have two options to get around this:
> 
> 1. Make `IVDataService::get` a regular instance method. Meaning you would have to use it the way shown above.
> 2. or, add the arguments in the constructor to `IVDataService::get`. So, you would have to pass `enable_cache=False` too `IVDataService::get` to disable the cache.
> 
> Im not really satisfied with either option, but I think the first is probably the better of the two IMO. Thoughts?

I can see arguments for both and both of them match design patterns we've employed on other tools. The first choice seems nice because advanced users that don't want caching may prefer to set `enable_cache` once. On the other hand `hdf_cache` works at the `get` level like the second option. I don't really have any strong feelings. I'm partial to the second option as it seems a bit less "stateless" and easier to manage if for some reason you want to switch between caching and not caching in the same process.

It's your baby, so I'll yield to your judgement. I just want it to work.

aaraney:
> I can see arguments for both and both of them match design patterns we've employed on other tools. The first choice seems nice because advanced users that don't want caching may prefer to set `enable_cache` once. On the other hand `hdf_cache` works at the `get` level like the second option. I don't really have any strong feelings. I'm partial to the second option as it seems a bit less "stateless" and easier to manage if for some reason you want to switch between caching and not caching in the same process.
> 
> It's your baby, so I'll yield to your judgement. I just want it to work.

I agree with both your points. As a user, I think both are valid use cases. As a developer, to your point, I think the second will be easier to maintain and avoid having a giant state machine. Below I drafted up a potential solution to overload `get`. It works, but it places a lot of assumptions on this code and any code that would be added.

```python
import pytest

class a:
    def __init__(self, some_bool=False) -> None:
        self.some_bool = some_bool

    def bar(*self, some_bool=False):
        if len(self) > 1:
            ValueError("too many variadic args")

        if self:
            print("im an instance method")
            inst = self[0]

            # handle state machine

        else:
            print("im a class method")
            inst = a(some_bool=False)

        print(f"\ninst: {id(inst)}")
        print(f"inst: {type(inst)}")
        print(f"inst: {inst}\n")


def test_a():
    inst = a()
    inst.bar()

    a.bar()

>>> im an instance method
>>> inst: 4376130224
>>> inst: <class 'test_delete_soon.a'>
>>> inst: <test_delete_soon.a object at 0x104d672b0>
>>> 
>>> im a class method
>>> inst: 4376130320
>>> inst: <class 'test_delete_soon.a'>
>>> inst: <test_delete_soon.a object at 0x104d67310>
```

aaraney:
After an morning of deliberation, I have decided to go with option 1, make `get` an instance method. This means that you can no longer do: `IVDataService.get(...)`, it is required to create an instance of the class -> `IVDataService().get(...)`. I think it's also appropriate to expose a `cache_enabled` property to the user that can be checked to determine if an instance of `IVDataService` has cache enabled or not.

For historical purposes, below is a potential decorator that will treat an instance method as an instance method or like class method. It is pretty hacky and has a lot of assumptions baked in, but ill leave it here for thought anyway:


```python
import inspect
from inspect import _ParameterKind as ParameterKind
from inspect import _empty as Empty
from functools import wraps


def wrap(o, implicit_callable):
    def outer(func):
        @wraps(func)
        def inner(*args, **kwargs):
            klass = globals()[o]
            signature = inspect.signature(klass.__init__)
            inst_args = signature.parameters

            accepted_kinds = [
                ParameterKind.POSITIONAL_ONLY,
                ParameterKind.VAR_POSITIONAL,
                ParameterKind.VAR_KEYWORD,
            ]

            error_message = f"Object constructor's with {' or '.join(map(lambda x: x.name, accepted_kinds))} args not supported by decorator."
            for parameter in inst_args.values():
                if isinstance(parameter.default, Empty) and parameter.name != "self":
                    raise ValueError(
                        "All non-`self` constructor args must have default values to use this decorator."
                    )
                if parameter.kind in accepted_kinds:
                    raise ValueError(error_message)

            # handle state machine
            if args:
                inst = args[0]

                # case: inst assumes role of `self`
                if isinstance(inst, klass):
                    args = args[1:]

                    # Assume `None` is sentinel value.
                    # replace `None` values with instance attribute reference
                    # note: a new instance of klass is created even if one is passed in.
                    # properties in the passed class that are present in the constructor
                    # are assumed to have the same name (i.e. no leading underscore).
                    # if those criteria are met, properties/entities provided in kwargs
                    # are passed to the constructor in place of default or prior present
                    # values
                    constructor_kwargs = {
                        k: getattr(inst, k) if kwargs.get(k) is None else kwargs.get(k)
                        for k in inst_args.keys()
                        if k != "self"
                    }

                # case: inst is not `self`
                else:
                    constructor_kwargs = {
                        k: v.default if kwargs.get(k) is None else kwargs.get(k)
                        for k, v in inst_args.items()
                        if k != "self"
                    }

                inst = klass(**constructor_kwargs)

            # guaranteed no args
            else:
                inst = klass(**kwargs)

            f = getattr(inst, implicit_callable, None)

            return f(*args, **kwargs)

        return inner

    return outer


class a:
    def __init__(self, some_bool=False) -> None:
        self.some_bool = some_bool

    @wrap("a", "boo")
    def bar(self, *, some_bool=False):
        ...

    def boo(self, *args, **kwargs):
        print(self.some_bool)
```

python/nwis_client/setup.py
```diff
@@ -28,11 +28,12 @@
 REQUIREMENTS = [
     "pandas",
     "numpy",
-    "requests",
```
jarq6c, line 31: Is `requests` still required? I had to install it to run.

python/nwis_client/hydrotools/nwis_client/iv.py
```diff
@@ -12,58 +12,113 @@
 
 import datetime
 from collections.abc import Iterable
-from multiprocessing import Pool
 from functools import partial
-from typing import List, Tuple, Union, Iterable
+from typing import Dict, List, Set, T, Union, Iterable
 import re
 import requests
```
jarq6c, line 18: Here it is.

python/nwis_client/hydrotools/nwis_client/iv.py
```diff
@@ -12,58 +12,113 @@
...
-        Max number of, per request retries, default 3
+    enable_cache : bool
+        Toggle sqlite3 request caching
+    cache_expire_after : int
+        Cached item life length in seconds
+
+    Examples
+    --------
+    >>> from hydrotools.nwis_client import IVDataService
+    >>> service = IVDataService()
```
jarq6c, line 45: Note: The `README.md` example will need to be updated since it currently calls `IVDataService.get` directly as a class method.

jarq6c (changes requested):
I've confirmed it is indeed significantly faster. The example below decreased from 2m43s to 0m44s. Retrieving all cached data took 0m16s. This was ~500,000 values from NWIS. In both cases, the gcp data was already cached and took less than 1s to load and process. So, raw retrieval from NWIS was taking ~28 seconds which in this case was ~70% decrease in retrieval time. Nice!

Disabling the cache also appeared to be functional.

It seems requests is still a dependency for `nwis_client.iv` and needs to be added back to the `setup.py`. Other than that, it passed all tests.

```python
# Import the GCP Client
from hydrotools.gcp_client import gcp

# Import the NWIS IV Client
from hydrotools.nwis_client.iv import IVDataService
import pandas as pd

# Instantiate model data service
model_data_service = gcp.NWMDataService()

# Retrieve forecast data
#  By default, only retrieves data at USGS gaging sites in
#  CONUS that are used for model assimilation
forecast_data = model_data_service.get(
    configuration = "short_range",
    reference_time = "20210101T01Z"
    )

# Validate site codes
valid_site_code = forecast_data['usgs_site_code'].str.isdigit()
forecast_data = forecast_data[valid_site_code]
forecast_data.loc[:, 'usgs_site_code'] = forecast_data['usgs_site_code'].cat.remove_unused_categories()

# Look at the data
print(forecast_data.info(memory_usage='deep'))
print(forecast_data[['value_time', 'value']].head())

# Generate site list
site_list = sorted(
    forecast_data['usgs_site_code'].cat.categories
)

# Extract period of observation
startDT = forecast_data['value_time'].min()-pd.Timedelta("1H")
endDT = forecast_data['value_time'].max()+pd.Timedelta("1H")

# Instantiate service
obs_service = IVDataService()

# Retrieve data from a single site
observations_data = obs_service.get(
    sites=site_list, 
    startDT=startDT, 
    endDT=endDT
    )

# Look at the data
print(observations_data.head())
```

aaraney:
> I've confirmed it is indeed significantly faster. The example below decreased from 2m43s to 0m44s. Retrieving all cached data took 0m16s. This was ~500,000 values from NWIS. In both cases, the gcp data was already cached and took less than 1s to load and process. So, raw retrieval from NWIS was taking ~28 seconds which in this case was ~70% decrease in retrieval time. Nice!

Wow! That's better than I would have expected! It might be worth adding that to the hydrotools readme just for advertising purposes. Thanks for doing some testing with it! Im stoked to see that result.

python/nwis_client/setup.py
aaraney: It was for type hints. I removed the import in `iv.py` and changed the type hints to their appropriate type, thus resolving this one. Nice catch, I need to be better about my dev testing environment.

python/nwis_client/hydrotools/nwis_client/iv.py
aaraney: I think I already got everything with fc552637938fb997d8ad9703c426aecd46df6da9, am I missing something else?

python/nwis_client/hydrotools/nwis_client/iv.py
```diff
@@ -438,157 +334,86 @@ def _get_raw_params_handler(
...
-            None,
-        ] = None,
-        period: Union[str, None] = None,
-        siteStatus: str = "all",
-        **params,
-    ):
-        params = self._get_raw_params_handler(
-            parameterCd, startDT, endDT, period, siteStatus, **params
-        )
+        filtered_members = [
```
aaraney, line 388: @jarq6c, it's worth noting here, in the future it would be pretty trivial to accept multiple major filters (sites, hucs, etc.). I don't know if that is desired as a feature, but I just wanted to mention it.

jarq6c (approved):

jarq6c:
Successfully installed, ran the above example, and passed all tests on my end.

PR #91, opened 2021-06-15
Pre-release nwis_client require  _restclient 3.0.0 and gh-actions install _restclient explicitly

aaraney:
Add `_restclient` version requirement for `nwis_client`. Now requires `3.0.0`

The Run Unit Tests `gh-action` now installs `_restclient` explicitly prior to installing the namespace package. See below for context.

## Notes

- This is a pre-release PR before the `restclient_transition` branch is merged into `main`

## Checklist

- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### DIFF
--- a/.github/workflows/run_unit_tests.yml
+++ b/.github/workflows/run_unit_tests.yml
@@ -22,6 +22,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip pytest pytest-aiohttp coverage coverage-badge
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        # see #91 for reasoning behind explicit install of _restclient
+        pip install -e "python/_restclient[develop]"
         python setup.py develop
     - name: Run all unittests 
       run: |
--- a/python/nwis_client/setup.py
+++ b/python/nwis_client/setup.py
@@ -28,7 +28,7 @@ DESCRIPTION = (
 REQUIREMENTS = [
     "pandas",
     "numpy",
-    "hydrotools._restclient",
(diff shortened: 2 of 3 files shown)


aaraney:
The tests fail because of the installation order. I tested this locally and I was able to reproduce the failing behavior of the test by installing using: `python setup.py develop`. However, if I install `_restclient` first with `pip install python/_restclient` then `pip install python/nwis_client` everything works well.

aaraney:
I see two ways to get around this problem:

1. add complexity to the namespace level `setup.py` by pushing packages that fail to install to some collection, install the rest of the packages  and retry to install the packages that failed. This is by no means a perfect solution and really only works if the dependency graph is 1 level deep. Below is an example where this quickly breaks down:

```shell
# -> means depends on
a -> b
b -> c
c
```

Both `a` and `b` would fail to install and the order of re-installation would determine success -- not optimal.

2. The most pragmatic solution to solve the problem of install dependencies in order in this case seems to just explicitly install `_restclient` in the `gh-action`. Its not an "optimal" solution, but given that a few packages depend on `_restclient` I think this is the best solution. A non-local version (`pypi` or GitHub`) of the dependent package also should not be installed as long as the specified minimum requirement for the dependency is `<=` the local version (`gh-action` or local dev env) of the dependent package.

aaraney:
Looks like 577e9ff resolved the issue. Given the 577e9ff addition changes the scope of this PR, I am going to change the title and description accordingly.

PR #97, opened 2021-06-17
Add long description and classifiers to _restclient and nwis_client setup.py

jarq6c:
Adds long descriptions and classifiers to bring `_restclient` and `nwis_client` up to parity with other subpackages.

### DIFF
--- a/python/_restclient/setup.py
+++ b/python/_restclient/setup.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 from setuptools import setup, find_namespace_packages
+from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


jarq6c:
All tests pass and pushing it through. FYI: got a little out of scope. @aaraney 

Got a deprecation warning on line 673 of `iv.py` because python mistook `\d` as a unicode character in the regex expression. The fix was to make the regex a raw string.

aaraney:
@jarq6c thanks for the bug fix and add the long descriptions, I'd forgotten to do that yesterday.

PR #138, opened 2021-10-01
Redesign NWM Client Subpackage

jarq6c:
This is a significant refactor and redesign of `nwm_client`. It implements all existing functionality and compatibility with Google Cloud Platform and generic http servers like NOMADS. The design is so different that I've added it to the repository as a new subpackage under `nwm_client_new`. The plan would be to transition the old `nwm_client` to this one after some further testing "in the field". This new package adopts a more component based design spread across 5 modules:

`NWMClient`: Top level interface responsible for corralling the other four components.
`NWMFileCatalog`: Interfaces to GCP and HTTP servers used to discover files based on simple queries.
`FileDownloader`: Asynchronous file downloader.
`NWMFileProcessor`: Processes raw NetCDF files to datasets and dataframes.
`ParquetCache`: Implements a parquet version of `HDFCache` to store processed dataframes.

As the tool is based on `dask`, I expect it to scale better than the existing client tool. However, per dask's best practices, the tool assumes most requests will fit in memory and therefore defaults to retrieving `pandas.DataFrame` with a simple switch to retrieve `dask.dataframe.DataFrame` for larger-than-memory datasets.

# Example usage
```python
# Import the file client
from hydrotools.nwm_client.NWMClient import NWMFileClient

# Setup the client, defaults to Google Cloud Platform
#  Note: Retrieval defaults to only those locations that correspond to usgs gage locations,
#  according to the NWM RouteLink files. You can specify a different mapping (also called a crosswalk)
#  to retrieve a custom subset
client = NWMFileClient(
#    location_metadata_mapping=my_custom_crosswalk
)

# Get a pandas.DataFrame
#  This method includes file retrieval, processing, and caching to parquet
df = client.get(
    configuration="analysis_assim",
    reference_times=["20210930T00Z"]
)

# Alternatively you can get a dask.dataframe.DataFrame by
#  turning off the dask compute option
#  In both cases, an HT canonical dataframe is returned
ddf = client.get(
    configuration="analysis_assim",
    reference_times=["20210930T00Z"],
    compute=False
)

# Get data from NOMADS
# Import the HTTPFileCatalog for use with generic web servers
# This has been successfully tested against NOMADS and a simple python3 -m http.server
from hydrotools.nwm_client.NWMFileCatalog import HTTPFileCatalog

# Setup the catalog
#  Note: If you have a custom CA Bundle, you would also pass in 
#  a custom ssl.SSLContext
NOMADS_catalog = HTTPFileCatalog(
    server="https://nomads.ncep.noaa.gov/pub/data/nccf/com/nwm/prod/",
#    ssl_context=my_custom_ssl_context
)

# Setup the client as before, but this time specify a catalog
NOMADS_client = NWMFileClient(
    catalog=NOMADS_catalog,
#    ssl_context=my_custom_ssl_context
)

# Get a pandas.DataFrame from NOMADS
df2 = NOMADS_client.get(
    configuration="analysis_assim",
    reference_times=["20210930T01Z"]
)
```

Inviting @aaraney @hellkite500 @christophertubbs to review.

## Testing

1. Each component is minimally tested for all functionality. I did not duplicate every test in the original `nwm_client` because many of those tests were redundant.

## Notes

- The primary interface takes a list of reference times, which are processed serially. This is purposeful, since each reference time may require hundreds of files to be downloaded and processed. The maximum number of simultaneous connections is also limited to 10 by default. GCP in particular did not test well as the number of simultaneous downloads approached 20. YMMV.
- The actual processing of each individual model cycle is performed in parallel by `dask`.
- To my knowledge, no current analyses involve all of CONUS across multiple cycles. This use case may require special attention. This tool would certainly make such an analysis easier, but as-is the primary interface assumes users will typically want a subset.

## Todos

- Right now the processor defaults to dask partitions that contain 2.4 million values. Based on back of the envelope calculations this should result in partitions just under 100 MB in size in memory. This is per dask's best practices. You can explicitly specify a partition size when you repartition, but this is computationally expensive. I left a TODO to make the number of partitions a selectable parameter that could be changed by the caller and further research may be required to find a good sized partition.

## Checklist


python/nwm_client_new/src/hydrotools/nwm_client/NWMClient.py
```diff
@@ -0,0 +1,379 @@
...
+        # Convert to dataframe
+        df = NWMFileProcessor.convert_to_dask_dataframe(ds)
+
+        # Canonicalize
+        return NWMClient.canonicalize_dataframe(
+            df=df,
+            configuration=configuration
+        )
+
+    def get(
```
hellkite500, line 251: It is not real clear (by reading the code at least) what the difference between `get_cycle` and `get` is, and it seems they are implemented in different ways.  Intuitively, I would think `get_cycle` would make multiple calls to `get`???  Also, why doesn't `get_cycle` attempt any caching or cleanup where `get` does?

python/nwm_client_new/src/hydrotools/nwm_client/NWMFileProcessor.py
```diff
@@ -0,0 +1,125 @@
...
+        # Open dataset
+        ds = xr.open_mfdataset(file_list, engine="netcdf4")
+
+        # Prepare feature_id filter
+        if len(feature_id_filter) != 0:
+            # Convert to array
+            feature_id_filter = np.asarray(feature_id_filter)
+
+            # Validate filter IDs
+            check = np.isin(feature_id_filter, ds.feature_id)
```
hellkite500, line 66: Is it worth simply `try`ing to select the filter and then catching a `KeyError` before doing the `check = np.isin(...)`.  This way you WARN about which ID's were not selectable

python/nwm_client_new/src/hydrotools/nwm_client/NWMFileProcessor.py
```diff
@@ -0,0 +1,125 @@
...
+        -------
+        dask.dataframe.DataFrame of NWM data.
+        """
+        # Convert to dask dataframe
+        df = ds.to_dask_dataframe()
+
+        # Compute number of partitions
+        #  Best practice is ~100 MB per partition
+        #  TODO Make this a parameter
+        npartitions = 1 + len(df.index) // 2_400_000
```
hellkite500, line 96: what is going on here `// 2_400_000`. haha.  Haven't seen `2_400_000` this kind of syntax before.

python/nwm_client_new/src/hydrotools/nwm_client/NWMFileProcessor.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Sort by feature ID
+        #  Most applications will benefit from this
+        df = df.sort_values(by="feature_id")
+
+        # Reset index
+        df = df.set_index("feature_id").reset_index()
```
hellkite500, line 103: What is this intended to do?  It sets the index to `feature_id` but then reindexes with integer indices.  Why?

python/nwm_client_new/src/hydrotools/nwm_client/ParquetCache.py
```diff
@@ -0,0 +1,104 @@
...
+        subdirectory: str
+            subdirectory used to reference object in parquet root directory.
+        *args
+            Additional arguments passed directly to function.
+        **kwargs
+            Additional arguments passed directly to function.
+
+        Returns
+        -------
+        object
```
hellkite500, line 70: Returns `DaskDataframe` according to type hint

python/nwm_client_new/src/hydrotools/nwm_client/data/routelink_files/RouteLink_PuertoRico_NWMv2.1_20191204.csv
```diff
@@ -0,0 +1,105 @@
+# Convention: CF-1.6
```
hellkite500, line 1: Might want to consider a different source of this data (not static in the repo) at some point.

python/nwm_client_new/tests/test_NWMClient.py
```diff
@@ -0,0 +1,54 @@
+import pytest
```
hellkite500, line 1: should probably test the client `get_cycle` method at least once somewhere.

hellkite500 (approved):
Don't see anything that would prohibit a merge, but there may be some cleanup that you want to do before.

PR #156, opened 2021-11-09
Switch NWM Crosswalk Source from CSV to URL

### LINKED ISSUE #145: Change crosswalk source to HydroShare
Currently `nwm_client_new` uses the old RouteLink CSVs to generate the default crosswalk. We want to change this to pull the crosswalk from HydroShare and cache it like any other dataframe.

jarq6c:
This PR changes the source of default National Water Model crosswalk information from packaged CSVs to an HDF5 file hosted on HydroShare.

https://www.hydroshare.org/resource/d154f19f762c4ee9b74be55f504325d3/

The crosswalk is no longer built on import. I implemented the crosswalk as a property of `NWMClientDefaults`. The interaction of this property with the `NWMClient` objects means the file is only downloaded and processed if a `NWMClient` is initialized without a user-specified `location_metadata_mapping`. The file is then downloaded via the `FileDownloader` and cached in the default `ParquetCache` like any other dataframe associated with the client.

## Additions

- `NWMClientDefaults.ROUTELINK_URL` specifies the location of the NWM RouteLink HDF5 file.

## Removals

- `data` directory and all RouteLink CSVs have been removed.

## Changes

- `NWMClientDefaults.CROSSWALK` is now a property that downloads and caches remote RouteLink data when called.

## Testing

1. Passes all existing tests.
2. No new tests. Enhancement uses existing components.

## Notes

- HydroShare seems quick enough for this small amount of data.
- Bandwidth limited users that cannot retrieve the crosswalk are unlikely to have success retrieving NWM data.
- Requesting review from @aaraney 
- Closes #145 

## Todos

- None

## Checklist


python/nwm_client_new/src/hydrotools/nwm_client_new/NWMClient.py
```diff
@@ -60,20 +63,53 @@ class NWMClientDefaults:
...
+                ssl_context=ssl_context
+                )
+
+            # Download files
+            downloader.get([(url, "RouteLink.h5")])
+            return dd.from_pandas(pd.read_hdf(Path(td)/"RouteLink.h5"), 
+                npartitions=1)
+    
+    @property
+    def CROSSWALK(self) -> pd.DataFrame():
```
aaraney, line 102: Is the call to `pd.DataFrame()` desired? Ive never seen this style of type hint, so I am more curious than anything.

python/nwm_client_new/src/hydrotools/nwm_client_new/NWMClient.py
```diff
@@ -60,20 +63,53 @@ class NWMClientDefaults:
...
-            rl_file,
-            dtype={"nwm_feature_id": int, "usgs_site_code": str},
-            comment='#'
-        ).set_index('nwm_feature_id')[['usgs_site_code']])
-        self.CROSSWALK = pd.concat(dfs)
+    ROUTELINK_URL: str = "https://www.hydroshare.org/resource/d154f19f762c4ee9b74be55f504325d3/data/contents/RouteLink.h5"
+
+    def get_routelink(
+        self, 
+        url: str,
```
aaraney, line 70: I have a few comments/questions about this method:

1. since this method does not use anything from `self`, it might be better suited as a `staticmethod`.
2. I like the flexibility of the `url` parameter, but why not also have flexibility for the output file name? 
3. IMO, given that this method serves a single purpose (kudos!), it might make sense to give defaults for the keyword arguments (`url=NWMClientDefaults.ROUTELINK_URL`, and `output_name=NWMClientDefaults.ROUTELINK_OUTPUT_NAME` or something like that)

python/nwm_client_new/src/hydrotools/nwm_client_new/NWMClient.py
aaraney: Given that we might want to support versions of the crosswalk in the future, it might be a good idea to support that. So, maybe something like move the innards the `CROSSWALK` property to a `get_crosswalk` method that takes in a version number (enum of versions we support) and returns that crosswalk. It would be great to then tag the dataframe with metadata (maybe `df.attrs["version"] = "2.1"`. `attrs`' is still not "stable", im just spitballing). Then the `CROSSWALK` property just returns the latest crosswalk. thoughts?

aaraney (commented):
In general this looks good! I have a few comments worth (i think) worth discussing. So, im looking forward to hearing back from you @jarq6c! Cheers.

python/nwm_client_new/src/hydrotools/nwm_client_new/NWMClient.py
jarq6c: That type hint is an error. I'm kind of surprised that works.

I agree with versioning in principle, but I'm not sure how achievable it is in practice. You already mentioned `attrs` still being unstable, but we also have a limitation on how NWM treats "versions." For example, version 2.2 of the NWM is set to release soon, but there will be no version 2.2 of auxiliary files. NWM 2.2 will be backwards compatible with all auxiliary files including RouteLinks.

There's an additional complication in that the client tools is compatible with the 2.0 NWM data on Google Cloud, even though it uses the NWM 2.1 RouteLink information. The distinction between NWM versions and NWM RouteLink versions is muddy.

I've added this topic as an agenda item for our next meeting. This might be a question we can pose to the Next Gen NWM team.

python/nwm_client_new/src/hydrotools/nwm_client_new/NWMClient.py
jarq6c: I'm mulling this over. I'm not sure these three recommendations are compatible with each other.

jarq6c: OK, it may not be what you had in mind, but I decided to err on the side of maintainability. I decided to rename `get_routelink` to `_download_and_read_routelink_file` and remove all parameters (except `self`). I think this is more suitable for an object that houses `NWMClientDefaults`. These values aren't really expected to be changed or used directly by a user (and should probably be spun off to their own module). In this vein, `NWMFileClient` should probably also be it's own module, but I'll leave these organizational concerns to a separate PR.

I initially thought about incorporating more methods to facilitate download and consumption of RouteLink files, but I think that's outside the scope of `NWMClientDefaults`. I'm also not sure what added functionality we could include. A preprocessed `pandas.DataFrame` is easily retrievable direct from the HDF5 file on HydroShare.

python/nwm_client_new/src/hydrotools/nwm_client_new/NWMClient.py
aaraney: Right, model versioning does not necessarily equal to parameter versioning. Ill close this and we can discuss this in the upcoming meeting. I dont think we want to become the source of "truth" for the mapping of nwm version's to parameter version's, hopefully we can find another source where that is documented and point people there and/or mirror that "truth".

python/nwm_client_new/src/hydrotools/nwm_client_new/NWMClient.py
aaraney: >  I decided to rename `get_routelink` to `_download_and_read_routelink_file` and remove all parameters (except `self`). I think this is more suitable for an object that houses `NWMClientDefaults`. These values aren't really expected to be changed or used directly by a user (and should probably be spun off to their own module). In this vein, `NWMFileClient` should probably also be it's own module, but I'll leave these organizational concerns to a separate PR.

👍🏻🚀

> I initially thought about incorporating more methods to facilitate download and consumption of RouteLink files, but I think that's outside the scope of `NWMClientDefaults`.

I agree with your judgment. @christophertubbs might have a NOMADs crawler that we could incorporate in the future to do route link or parameter file "discovery", although I am not so keen on the idea. Just a thought though.

jarq6c:
Resolved all comments. Passes all tests. Thanks for the review @aaraney !

PR #158, opened 2021-11-15
NWM Client: Limit number of get calls for testing

jarq6c:
This update just limits the number of times tests for this subpackage make a request for NWM data. Slow tests would sometimes fail. I suspect the issue was lack was resources on the GH side. Previously, the tests ran all configurations, but that was excessive and did not increase our robustness since the same code was just tested repeatedly with similar circumstances. Specific fail-cases are already tested elsewhere.


jarq6c:
Note: I reran the slow tests locally for Python 3.7 and they succeeded.

PR #174, opened 2022-02-07
Fix 160. **params arg in IVDataService.get now raises warning when case insensitively matches other parameters

### LINKED ISSUE #160: NWIS Client: Possible case sensitivity issue
It appears the NWIS IV service is case-insensitive. For example, note the use of `startDT` and `startdt` in the two calls below:

https://waterservices.usgs.gov/nwis/iv/?format=json&sites=01646500&startDT=2021-11-01&endDT=2021-11-02&parameterCd=00060&siteStatus=all

https://waterservices.usgs.gov/nwis/iv/?format=json&sites=01646500&startdt=2021-11-01&enddt=2021-11-02&parameterCd=00060&siteStatus=all

This may be an issue for us because I think the current `nwis_client` validation is case-sensitive. This has caused some issues when using non-string datetime parameters. For example:

```python
client = IVDataService()

# Ex1. This succeeds
obs = client.get(sites="01013500", startdt="2021-09-01T00:00Z")

# Ex2. This fails
obs = client.get(sites="01013500", startdt=pd.Timestamp("2021-09-01T00:00Z"))

# Ex.3 This suceeds
obs = client.get(sites="01013500", startDT=pd.Timestamp("2021-09-01T00:00Z"))
```
I'm hesitant to have the validator transform any arguments unexpectedly. However, it might be a good idea for the date converter/validator to be case-insensitive. So, if a user enters `StArTdT` as a parameter the key is passed onto NWIS un-altered, but the value is still treated the same way we currently treat `startDT`.

ping @aaraney

aaraney:
`IVDataService.get`'s `**params` argument is susceptible to malformed case entries (i.e. `service.get(startdt="2022-01-01", ...)` should instead be `service.get(startDT="2022-01-01", ...)`). This PR resolves this issue. Now, the above case, or any case where a `**params` argument case insensitively matches a non-`**params` argument, throws a `RuntimeWarning` and notes the parameters that are likely malformed. This behavior should change to an exception once `nwis_client` > 3.0.X. 

fixes #160.

## Changes

- `nwis_client.IVDataService.get` throws `RuntimeWarning` when arguments in `**params` case insensitively match a non-`**params` argument. (i.e. `service.get(startdt="2022-01-01", ...)` the correct call is `service.get(startDT="2022-01-01", ...)`)

## Todos

- Throw `RuntimeError` once `nwis_client` > 3.0.X

## Checklist


python/nwis_client/tests/test_nwis.py
```diff
@@ -433,7 +446,19 @@ def get_raw_mock(*args, **kwargs):
...
+
+def test_nwis_client_get_throws_warning_for_kwargs(mocked_iv):
+    from packaging import version
+    version = version.parse(nwis_client.__version__)
+    version = (version.major, version.minor)
+
+    # versions <= than 3.1 should throw an exception instead of a warning
+    assert version <= (3, 1)
+
+    with pytest.warns():
```
jarq6c, line 462: I think you can specify the type of warning like this `pytest.warns(RuntimeWarning)`. I've also just learned you can do something like this to check the warning in more detail:

```python
with pytest.warns(RuntimeWarning) as record:
    warnings.warn("My RuntimeWarning Message", RuntimeWarning)

    assert len(record) == 1
    assert record[0].message.args[0] == "My RuntimeWarning Message"
```

jarq6c:
I like the plan to transition from warnings to exceptions. I think using the decorator was a good idea to facilitate this.

Thanks for the contribution!

jarq6c (approved):

aaraney:
See #176:

> ## Changes
> 
>     * alteration to goal set out in [174](https://github.com/NOAA-OWP/hydrotools/pull/174). Now, **`nwis_client` > 3.1** will raise `RuntimeError` exception instead of `RuntimeWarning` when arguments in `**params` case insensitively match a non-`**params` argument. (i.e. `service.get(startdt="2022-01-01", ...)` the correct call is `service.get(startDT="2022-01-01", ...)`). Previously, the proposed warning to error timeline was **`nwis_client` > 3.0**.

PR #184, opened 2022-03-03
Add check for complete categories when constructing contingency tables

### LINKED ISSUE #183: contingency_table computed from compute_contingency_table doesn’t contain both True and False for the observed or the simulated series
The following line in function compute_contingency_table:

ctab = pd.crosstab(observed, simulated, dropna=False) 
produces ctab containing only True or False for either the observed or the simulated series? It would cause the error in ctab.loc. 

Could you add a solution to handle such situation?

jarq6c:
This adds a check and corrects missing boolean categories in Categorical series input to `compute_contingency_table`.

## Additions

- Checks for both `True` and `False` in observed and simulated input series.

## Removals

- None

## Changes

- Update original test to use categorical series instead of raw categories per the original docstring and intended use of `compute_contingency_table`.

## Testing

1. Four new test scenarios for `compute_contingency_table` that check for all `True` and all `False` cases.

## Notes

- Fixes #183 

## Todos

- None

## Checklist

### DIFF
--- a/python/metrics/src/hydrotools/metrics/metrics.py
+++ b/python/metrics/src/hydrotools/metrics/metrics.py
@@ -196,9 +196,9 @@ def compute_contingency_table(
     Parameters
     ----------
     observed: array-like, required
-        pandas.Series of boolean pandas.Categorical values indicating observed occurrences
+        Array-like of boolean values indicating observed occurrences
     simulated: array-like, required
-        pandas.Series of boolean pandas.Categorical values indicating simulated occurrences
+        Array-like of boolean values indicating simulated occurrences
     true_positive_key: str, optional, default 'true_positive'
         Label to use for true positives.
     false_positive_key: str, optional, default 'false_positive'


jarq6c:
Tagging @aaraney

jarq6c:
@hellkite500 @aaraney  any feedback on this PR?

python/metrics/src/hydrotools/metrics/metrics.py
```diff
@@ -214,6 +214,16 @@ def compute_contingency_table(
         pandas.Series of integer values keyed to pandas.Index([true_positive_key, false_positive_key, false_negative_key, true_negative_key])
         
     """
+    # Check for complete list of categories
+    if True not in observed.cat.categories:
```
hellkite500, line 218: I'm not super familiar with categorical types...can you no use `any` or something like that? Or is `in` the preferred method of testing?

jarq6c: Ugh... this actually made me realize that that line doesn't actually do what I want (and neither will `any`). In this case, we need the behavior of `is`, but `in` works like `==`. I should probably separate this check out to another method and test it.

python/metrics/src/hydrotools/metrics/_validation.py
```diff
@@ -100,3 +103,65 @@ def raise_for_inconsistent_shapes(
...
+    no_false_category = True
+    for c in series.cat.categories:
+        if c is False:
+            no_false_category = False
+    if no_false_category:
+        message = f"{series} has no False category, attempting to add"
+        warnings.warn(message=message, category=UserWarning)
+        series = series.cat.add_categories([False])
+
+    # Check number of categories
```
hellkite500, line 162: Move this check above the type/inclusion checks...no point in testing what each one is if there are too many...

jarq6c: There's more to the story.

Here's a use-case example:
```python
# Say we have a flooding threshold we use to transform a continuous 
#  variable into a categorical variable
flood_threshold = 40.0

# Here's our continues variable (volumetric discharge)
flow_values = pd.Series([41.0, 42.0, 43.0])

# Here's where we convert the flow values into categorical values
observed_flood = (flow_values >= flood_threshold).astype("category")
```

`observed_flood` is a categorical `Series` with values `[True, True, True]` and categories `[True]`. This causes an error when attempting to cross-tabulate when generating contingency tables because it does not contain a `False` category. I was trying to implement a user friendly way to:

1. Add `True/False` to the available categories if missing
2. Make sure that if `True` or `False` were added as categories that there are only two categories

For example, if a categorical `Series` has categories `[False, "foo"]`, this would pass because it has two categories. Then `True` would get added resulting in categories `[False, "foo", True]`. This would raise an error down the road. That's why I check for the number of categories last.

jarq6c: Refactor to force it to create a new Categorical Series.

python/metrics/src/hydrotools/metrics/_validation.py
```diff
@@ -100,3 +103,65 @@ def raise_for_inconsistent_shapes(
...
+        warnings.warn(message=message, category=UserWarning)
+        series = series.astype("category")
+
+    # Check for True category
+    no_true_category = True
+    for c in series.cat.categories:
+        if c is True:
+            no_true_category = False
+    if no_true_category:
+        message = f"{series} has no True category, attempting to add"
```
aaraney, line 148: what does this print? I would assume that it would print the, series and then `has no True category, attempting to add`. Is that right? Or am I missing something?

jarq6c: Correct, though this has been removed.

python/metrics/src/hydrotools/metrics/metrics.py
```diff
@@ -195,9 +195,9 @@ def compute_contingency_table(
         
     Parameters
     ----------
-    observed: pandas.Series, required
+    observed: array-like, required
```
hellkite500, line 198: why does this indicate an `array-like` type but the string below says `pandas.Series`?

hellkite500:
Nice work on the refactor, by the way!  Much simpler, and even more general!  I like it!

jarq6c:
> Nice work on the refactor, by the way! Much simpler, and even more general! I like it!

Thanks for the review!

hellkite500 (approved):

hellkite500:
Approved, but I'll give @aaraney a chance to look it over before merging.

jarq6c:
This has been published to PyPI.

aaraney:
sorry for being slow to get to this. having looked at it now, I would have approved the merge. like @hellkite500's noted, I think the refactor was done nicely. Nice work!

jarq6c:
> sorry for being slow to get to this. having looked at it now, I would have approved the merge. like @hellkite500's noted, I think the refactor was done nicely. Nice work!

Thanks @aaraney !

PR #189, opened 2022-04-18
Improvements to validation checks

groutr:
Some minor improvements to _validation.py that center around the usage of `np.asarray`. Using `np.asarray` over `np.array` can avoid unnecessary copies (ie, when the input is already an array).

When comparing the shapes, we don't need to hold all the arrays in memory.

### DIFF
--- a/python/metrics/src/hydrotools/metrics/_validation.py
+++ b/python/metrics/src/hydrotools/metrics/_validation.py
@@ -89,19 +89,15 @@ def raise_for_inconsistent_shapes(
     InconsistentShapesError:
         Raises if all of the arrays are not the same shape.
     """
-    # Convert to numpy arrays
-    arrays = [np.array(x) for x in arrays]
-
     # Extract first array
-    x = arrays[0]
+    xshape = np.asarray(arrays[0]).shape
 
     # Check shape of each
-    for y in arrays:
+    for y in map(np.asarray, arrays[1:]):
         # Test each array
-        test = all(i == j for i, j in zip(x.shape, y.shape))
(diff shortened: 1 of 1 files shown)


aaraney:
@groutr, thanks for the contribution!

I am not sure the `_array_attr` function is necessary. `np.asarray` does [not copy](https://numpy.org/doc/stable/reference/generated/numpy.asarray.html#numpy.asarray) if it receives an input `np.array`. I think it would be more clear to just cast to `np.asarray` and check for the property in places you are proposing to call this method.

aaraney:
@groutr, thanks for the contribution!

I am not sure the `_array_attr` function is necessary. `np.asarray` does [not copy](https://numpy.org/doc/stable/reference/generated/numpy.asarray.html#numpy.asarray) if it receives an input `np.array`. I think it would be more clear to just cast to `np.asarray` and check for the property in places you are proposing to call this method.

aaraney:
Ah, I see. That makes a lot of sense. Thanks for clarifying and providing the above example, @groutr.

My only request would be adding some type hints to the function declaration. The return type hint might be difficult to nail down for all cases, so I think the straight path of `npt.ArrayLike` should be fine unless you have a better suggestion.

jarq6c (changes requested):
Looks good to me. I think we just want to bump the patch version in `python/metrics/src/hydrotools/metrics/_version.py` to `1.2.2`

I tested these changes locally and it is indeed faster. Thanks for the contribution!

aaraney:
Thanks again, @groutr!

Issue #113, opened 2021-07-28
Add method to transform units returned by gcp_client and nwis_client


jarq6c:
So annoying to remember to transform units from gcp_client to compatible units with nwis_client.

jarq6c:
Closing old issues. TEEHR can handle unit conversions: https://github.com/RTIInternational/teehr

PR #196, opened 2022-06-02
Add minimal unit handling to `nwm_client`

### LINKED ISSUE #137: Investigate unit of measurement handling
HT currently has no formal unit of measurement handling. We may be able to incorporate a library like Pint to deal with this.

https://pint.readthedocs.io/en/stable/

jarq6c:
This PR adds minimal unit handling to `nwm_client`. This is an incremental improvement that includes some housekeeping. Future enhancements will unify the `gcp` and `http` interfaces similar to `nwm_client_new`, and transition from `google-cloud-storage` to `fsspec` libraries.

@aaraney  may be interested.

## Additions

- `unit_system` parameter. Setting this parameter to `"US"` will convert dataframe values returned by `get` to cubic feet per second and update `measurement_unit` to `"ft^3/s"`.
- `UnitHandler` class with methods to convert units.

## Removals

- The old deprecated `gcp_client` has been removed.

## Changes

- The `gcp` target for pip has been removed and `google-cloud-storage` is brought in automatically.

## Testing

1. Value conversions are tested in `test_units.py`
2. `pandas.Series` conversions are tested in `test_gcp.py`

## Notes

- Closes #137 
- Link to #113 

## Todos

- Unify gcp and http interfaces.

## Checklist

### DIFF
--- a/python/nwm_client/src/hydrotools/nwm_client/UnitHandler.py
+++ b/python/nwm_client/src/hydrotools/nwm_client/UnitHandler.py
@@ -73,10 +73,7 @@ class UnitHandler:
         """
         # Check for pandas.series
         if isinstance(value, pd.Series):
-            return pd.Series(
-                data=self.unit_registry.Quantity(value.values, from_units).to(to_units).magnitude,
-                index=value.index
-            )
+            return self.unit_registry.Quantity(value.values, from_units).to(to_units).magnitude
 
         # Return converted value
         return self.unit_registry.Quantity(value, from_units).to(to_units).magnitude


jarq6c:
@hellkite500 sorry for the large number of files, I removed the `gcp_client` which is outdated. The main files with the new functionality are `UnitHandler.py`, `gcp.py`, and their associated test files.

python/nwm_client/src/hydrotools/nwm_client/UnitHandler.py
```diff
@@ -0,0 +1,82 @@
...
+            Converted values same shape as value.
+
+        Example
+        -------
+        >>> unit_handler = UnitHandler.UnitHandler()
+        >>> unit_handler.convert_values([1, 1, 1], "ft", "m")
+        array([0.3048, 0.3048, 0.3048])
+        """
+        # Check for pandas.series
+        if isinstance(value, pd.Series):
```
hellkite500, line 75: Does using an arraylike view here work?  If so, it would simplify this code and remove the `isinstance` check I think.
```python
converted = self.unit_registry.Quanitity(value[:], from_units).to(to_units).magnitude
value[:] = converted
return value
```
This would essentially do an inplace update and return the reference to the thing updated.  If you really wanted a copy, then the existing might be the simplest?

hellkite500 (approved):
Looks good as is.  Just had some thoughts on the implementation of `convert_values` to ponder.

python/nwm_client/src/hydrotools/nwm_client/UnitHandler.py
hellkite500: This works about the same as I was thinking, but does have a caveat of only really being applicable to values, because of the slice/view here:
```python
df.loc[:, "value"] = self._unit_handler.convert_values(df["value"], "m^3/s", "ft^3/s")
```
the values get updated, but something like this wouldn't work (which is what I thought you were looking for)
```python
s = pd.Series([1,1,1])
s_converted = convert_values(s, "m^3/s", "ft^3/s")
```
`s_converted` would be an array, not a series.
I wonder if it would be worth writing this in a way you could `apply` or `map` the function?
```python
s_converted = s.apply(convert_values, args=("m^3/s", "ft^3/s"))
```

Though I think what you have here is both useful and efficient for the time being.

hellkite500 (approved):

PR #212, opened 2022-11-29
Minor improvements

groutr:
A tiny collection of minor improvements.

### DIFF
--- a/python/events/src/hydrotools/events/event_detection/decomposition.py
+++ b/python/events/src/hydrotools/events/event_detection/decomposition.py
@@ -91,7 +91,7 @@ def detrend_streamflow(
         
         """
     # Check index
-    if type(series.index) != pd.DatetimeIndex:
+    if not isinstance(series.index, pd.DatetimeIndex):
         raise Exception("series index is not DatetimeIndex")
 
     if not series.index.is_monotonic_increasing:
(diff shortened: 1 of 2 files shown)


python/events/src/hydrotools/events/event_detection/decomposition.py
```diff
@@ -53,22 +54,12 @@ def rolling_minimum(
...
-        index=series.index)
-
-    # Estimate trend using a backward rolling minimum
-    backward = reverse.rolling(window=window).min()
-
-    # Restore reversed series
-    backward = pd.Series(backward.values[::-1], 
-            index=backward.index)
-
+    forward = series[::-1].rolling(window=window).min()[::-1]
```
jarq6c, line 57: @groutr I realize this review has come _way_ too late. However, I recently had time to revisit this PR and I tested this particular change. I can confirm that these changes generate an identical final `rolling_minimum` time series. However, at least on my system I found that these changes resulted in a ~12% slower execution than the original (using `timeit` averaged across 10,000 executions). I also found that these changes use about the same amount of peak memory (sometimes slightly more than the original).

I'm honestly flummoxed as to an explanation, because I honestly thought this change would be result in faster execution or less memory usage due to not shoving numpy views into new pandas series.

Care to speculate?

jarq6c:
Unable to verify performance improvements.

PR #237, opened 2023-09-19
Bump _restclient's `aiohttp` and `aiohttp_client_cache` version deps

aaraney:
bump `aiohttp` and `aiohttp_client_cache` version deps
    
See #103 for background as to why `aiohttp` was previously pinned. In short, the issues that led to pinning `aiohttp` have be resolved upstream.
    
`aiohttp_client_cache` >= 0.9.0 needed b.c. https://github.com/requests-cache/aiohttp-client-cache/issues/173

## Changes

- `_restclient` depends on `aiohttp~=3.8` and `aiohttp_client_cache>=0.9.0`

## Checklist

### DIFF
--- a/python/_restclient/pytest.ini
+++ b/python/_restclient/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+asyncio_mode = auto
--- a/python/_restclient/setup.cfg
+++ b/python/_restclient/setup.cfg
@@ -30,8 +30,8 @@ packages = find_namespace:
 package_dir =
     =src
 install_requires =
-    aiohttp<=3.7.4.post0
-    aiohttp_client_cache
+    aiohttp~=3.8
+    aiohttp_client_cache[sqlite]>=0.9.0
     python-forge
     aiosqlite
     pandas


jarq6c:
Simple enough!

aaraney:
I will open up another PR that bumps the version of this when i'm done testing things out. Im running into an issue with the `nwis_client` `cli`.

jarq6c:
Any idea why this PR would cause the `deploy-gh` workflow to fail?

aaraney:
I think it is because the source didnt change. Still kind of odd that it failed.

aaraney:
Having spend a day or so looking at this last week, I don't know that it is going to be as straight forward to fix as I was hoping. Here is a brief-ish explanation. 

`hydrotools._restclient` uses and expands upon `aiohttp_cache_client` to provide a request client that handles request caching and retry / backoff capabilities. `aiohttp_cache_client` supports several cache storage backends, `sqlite` being one of them. `sqlite` is the only backend that `_restclient` supports at the moment. `aiohttp_cache_client` uses `aiosqlite` as a backend driver to support its `sqlite` backend capabilities. `aiosqlite` use threading to provide a async sqlite client. To accomplish this, `aiosqlite` runs its own event loop (see [`aiosqlite.Connection.run()`](https://github.com/omnilib/aiosqlite/blob/40442815b1a9a67b6f458fe56a1f4717c6f6c74c/aiosqlite/core.py#L76)) in a separate thread that effectively acts like a scheduler. 

Inside `aiosqlite`'s loop, a tuple of a callable _and_ an unfulfilled `asyncio.Future` are pulled from a work queue. The callable is called and the result of the callable is _scheduled_ to be set _as the result_ of the unfulfilled `asyncio.Future` _on the **main** event loop_. This includes closing the `aiosqlite.Connection` instance. 

This means that the _runtime of the event loop_ must be strictly longer than the `aiosqlite.Connection` or else you cannot close the `aiosqlite.Connection` instance because the created future will never have a result. Likewise, this means that if a `aiosqlite.Connection` is created, but not explicitly or implicitly (`async` context manager) closed, the `aiosqlite.Connection` event loop will run forever and the program will not exit. Consequently, this means that any code that manages a `aiosqlite.Connection` instance is forced to require users of that code to explicitly or implicitly close the, possibly encapsulated, `aiosqlite.Connection`. Likely meaning that either a `close()` method is added to the public api of the object or `async` context manager is added and either is required to correctly use the object or the code will hang without exiting. A full description of the issue along with code examples can be found [here](https://github.com/requests-cache/aiohttp-client-cache/issues/173#issuecomment-1730206380).

aaraney:
@jarq6c, for sure! Yeah, I will open an issue to track this.

aaraney:
I opened #238 to track this.

PR #261, opened 2024-11-19
Update `hydrotools.metrics` to newer Python versions

jarq6c:
This PR updates the `hydrotools.metrics` package to test on Python versions 3.9, 3.10, and 3.11. It also transitions the package from using a `setup.cfg` to just using `pyproject.toml`.

## Additions

-

## Removals

- `setup.cfg`

## Changes

- `pyproject.toml`
- Had to switch some docstrings to raw strings due to embeded LaTeX.

## Testing

1. No changes.

## Notes

- Last version under old DOC license.
- Related to #259 and #260 

## Todos

-

## Checklist

- [x] PR has an informative and human-readable title
- [ x PR is well outlined and documented. See [#12](https://github.com/jarq6c/evaluation_tools/pull/12) for an example
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output) using [numpy docstring](https://numpydoc.readthedocs.io/en/latest/format.html) formatting
- [ ] Placeholder code is flagged / future todos are captured in comments
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #263, opened 2024-11-19
Migrate `setup.cfg` to `pyproject.toml`

aaraney:
- Update github actions imports `actions/setup-python` -> `@5`; `actions/checkout` -> `@v4`
- Test all subpackages in ci w/ `3.11` and `3.12`.
- Migrate all subpackage metadata files from `setup.cfg` to `pyproject.toml`.
- Bump all subpackage patch versions


aaraney:
@jarq6c, [`hydrotools.caches` tests](https://github.com/NOAA-OWP/hydrotools/actions/runs/11921332173/job/33225219879?pr=263) are failing for `python` `3.9`. It looks like a clean up condition that might need a guard clause. I verified that the later versions of python (`>3.9`) in the test matrix are using a later version, `3.10.1`, of `tables`. It looks like `tables` [3.10.0](https://github.com/PyTables/PyTables/releases/tag/v3.10.0) dropped support for `python<3.10`.

aaraney:
@jarq6c, I think the `hydrotools.nwm_client_new` `3.9` tests are failing for a similar reason. My best guess is the wheel pulled for `tables==3.9.2` (version in failing tests) was compiled with `numpy<2` but notice the test runner is pulling `numpy>=2`. So, there are ABI issues when `tables` goes to call `numpy` functions.

If that is the case, we should be able to remedy this by conditionally requiring `numpy<2` for `python<3.10`.

aaraney:
Looks like [9b9fcfa](https://github.com/NOAA-OWP/hydrotools/pull/263/commits/9b9fcfa3c2fad0a41d10c48325568087157bf947) fixed the `hydrotools.nwm_client_new` test failures.

aaraney:
Looks like [17c0090](https://github.com/NOAA-OWP/hydrotools/pull/263/commits/17c00903902003eb45b5d9377bece086a55ac48e) fixed the `hydrotools.caches` test failures.

python/nwis_client/pyproject.toml
```diff
@@ -1,6 +1,46 @@
 [build-system]
+requires = ["setuptools>=70"]
 build-backend = "setuptools.build_meta"
-requires = [
-  "setuptools>=42",
-  "wheel",
+
+[project]
+name = "hydrotools.nwis_client"
+authors = [{ name = "Austin Raney", email = "aaraney@protonmail.com" }]
```
jarq6c, line 7: It looks like the move to `pyproject.toml` makes it easier to have multiple authors. Would you like to share the authorship? You should remain first author, of course.

jarq6c:
Thanks for tackling all these @aaraney ! I had originally planned to bump the patch versions of all these packages with these changes. What do you think?

aaraney:
@jarq6c, yeah that is cool with me. Ill do that now.

python/metrics/src/hydrotools/metrics/_version.py
```diff
@@ -1 +1 @@
-__version__ = "1.3.5"
+__version__ = "1.3.6"
```
jarq6c, line 1: `metrics` was already bumped under #261

aaraney: Great catch, @jarq6c. Just reverted.

jarq6c (approved):

aaraney:
Thanks for the review, @jarq6c :tada:!

PR #264, opened 2024-11-20
ci: run slow unit tests on 3.11 & 3.12

### LINKED ISSUE #259: Update automated testing
Python 3.8 is EOL as of 2024-10-07. We need to update the automated tests and PyPI classifiers of all packages. I'm thinking we should target the three versions that are out of "bugfix" status and 3.12 (the oldest version still in bugfix).

aaraney:
closes #259

### DIFF
--- a/.github/workflows/run_slow_unit_tests.yml
+++ b/.github/workflows/run_slow_unit_tests.yml
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     steps:
     - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/run_slow_unit_tests.yml
```diff
@@ -87,7 +87,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
```
jarq6c, line 90: Good catch :)

.github/workflows/run_slow_unit_tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     steps:
     - uses: actions/checkout@v2
```
jarq6c, line 43: I think we'll want to update these, as well. (i.e. `actions/checkout` -> `@v4`)

aaraney: Thanks for catching this! Apparently need more coffee this morning 😅

jarq6c (approved):

jarq6c:
Thanks again @aaraney !

PR #268, opened 2024-12-03
Swap `nwm-client`

jarq6c:
Alright, let's rip this band-aid. This PR upgrades `nwm_client_new` v8.0.0, renames it, replaces the old `nwm_client` and removes the `nwm_client_new` package. v8.0.0 of both `nwm_client_new` and `nwm_client` have been deployed. The only difference apart from the package name is that `nwm_client_new` raises a `FutureWarning` on import.

This PR was always going to be messy due to the number of files edited and moved around.


jarq6c:
Looks like GH picked up the correctly named tests and they all passed!

aaraney:
Great! :tada:

PR #271, opened 2024-12-23
Remove caches and and old nwm client from slow unit tests

jarq6c:
Simple fix for failing slow unit tests that removes testing for `caches` and `nwm_client_new` which no longer exist.


jarq6c:
Manual run of weekly tests all passed.

PR #283, opened 2025-02-18
update pytest fixture from loop to event_loop in svi_client

jarq6c:
Weekly unit test failed yesterday. Same problem and fix addressed in #282


aaraney:
Thanks for fixing this, @jarq6c!

PR #307, opened 2026-05-01
Add `waterdata-client` dataframe transformers

jarq6c:
This PR introduces new functionality to transform deserialized GeoJSON into alternative formats. Prior to this PR, all classes defined in `clients.py` could only return `dict`. This PR adds the ability to further process these results into more useful formats for data scientists. This PR also adds column label mapping from WaterData labels to HydroTools canonical labels for some backwards compatibility with `hydrotools.nwis_client`.

## Changes

- `pyproject.toml` Added `geopandas` as a dependency.
- `base_client.py` Added new `transformer` attribute (a `Callable` type) and `_handle_response` method that calls the `transformer`.
- `clients.py` Regenerated from template.
- `constants.py` Regenerated from template.
- `transformers.py` New module that defines a `ResponseTransformer` protocol and `TransformedResponse_co` TypeVar. Also includes three example transformer methods: `check_features`, `to_geodataframe`, and `to_dataframe`.
- `clients.py.j2` Updates Jinja2 template for `clients.py` to include application of optional tranformer.
- `constants.py.j2` Updates Jinja2 template with column enums and mapping from waterdata labels to hydrotools canonical columns.
- `README.md` Added example using new `to_geodataframe` transformer.

## Testing

 - `test_clients.py` Updated to account for new transformer kwarg. Set to None to maintain current tests.
 - `test_transformers.py` Add testing for transformers.

## Checklist

### DIFF
--- a/python/waterdata_client/src/hydrotools/waterdata_client/base_client.py
+++ b/python/waterdata_client/src/hydrotools/waterdata_client/base_client.py
@@ -184,7 +184,7 @@ class BaseClient(Generic[TransformedResponse_co]):
     def _handle_response(
             self,
             data: list[dict[str, Any]]
-    ) -> TransformedResponse_co:
+    ) -> TransformedResponse_co | list[dict[str, Any]]:
         """Handle JSON response and optionally transform.
     
         Args:
@@ -195,5 +195,5 @@ class BaseClient(Generic[TransformedResponse_co]):
             transformed responses.
         """
         if self.transformer is None:
-            return data # type: ignore
+            return data
         return self.transformer(data)
--- a/python/waterdata_client/src/hydrotools/waterdata_client/base_client.py
+++ b/python/waterdata_client/src/hydrotools/waterdata_client/base_client.py
@@ -19,9 +19,9 @@ from .url_builder import (
     build_request_batch_from_feature_ids,
     QueryType
 )
-from .transformers import TransformedResponse_co, ResponseTransformer, check_features
+from .transformers import TransformedResponseT_co, ResponseTransformer, check_features
 
-class BaseClient(Generic[TransformedResponse_co]):
+class BaseClient(Generic[TransformedResponseT_co]):
     """Base class for USGS OGC API clients. Specific child classes may overwrite
     private attributes: _server, _api, _endpoint, _path, _content_type,
     _max_pages.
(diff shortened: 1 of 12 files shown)


python/waterdata_client/src/hydrotools/waterdata_client/base_client.py
```diff
@@ -174,3 +180,20 @@ def _get_next_url(self, response: dict[str, Any]) -> URL:
...
+    
+        Args:
+            data: A list of deserialized GeoJSON responses from an OGC-compliant API.
+        
+        Returns:
+            If self.transformer is None, returns untransformed data, else returns
+            transformed responses.
+        """
+        if self.transformer is None:
+            return data # type: ignore
```
christophertubbs: What happens if you add `list[dict[str, Any]]` to a union as the return type? It'll probably move the warning down the line, but it might be more insightful.

christophertubbs (approved):
It might have been better to set `TransformedResponse_co` to something like:

`TransformedResponse = Sequence[Mapping[str, Any]]`

in order to keep the interface uniform where the transformer's duties are just to modify what will eventually get passed into other constructors. May also be worth coming up with an alias for `list[dict[str, Any]]` anyways since it sounds like it fits `Rows` or something.

The only thing I consider a tad shaky is the vagueness of what `TransformedResponse_co` _might_ be, but I'm not looking at this in an editor and the possibility of IDE confusion doesn't qualify as a blocker.

python/waterdata_client/src/hydrotools/waterdata_client/base_client.py
jarq6c: Interesting. That seems to have fixed the issue. I swear when I tried adding it before it resulted in the incorrect type getting reported in `client.get`.

jarq6c:
> It might have been better to set `TransformedResponse_co` to something like:
> 
> `TransformedResponse = Sequence[Mapping[str, Any]]`
> 
> in order to keep the interface uniform where the transformer's duties are just to modify what will eventually get passed into other constructors. May also be worth coming up with an alias for `list[dict[str, Any]]` anyways since it sounds like it fits `Rows` or something.
> 
> The only thing I consider a tad shaky is the vagueness of what `TransformedResponse_co` _might_ be, but I'm not looking at this in an editor and the possibility of IDE confusion doesn't qualify as a blocker.

`TransformedResponse_co` is a template type variable. It could literally be anything. Here's a (rather silly) motivating example of what the type engine does with the TypeVar. In VSCode, the linter correctly shows that `client.get` returns `int`.

```python
from typing import Any
from hydrotools.waterdata_client import LatestContinuousClient

# Example custom transformer
def my_custom_transformer(data: list[dict[str, Any]]) -> int:
    return 42

# Instantiate client with custom transformer
client = LatestContinuousClient(
    transformer=my_custom_transformer
)

# Call client.get
value = client.get(
    monitoring_location_id="USGS-02146470"
    )

# Look at result
print(value) # 42
```

jarq6c:
Actually, I think I'll rename the type variable to `TransformedResponseT_co` to make it more explicit.

jarq6c:
WRT a type alias for `list[dict[str, Any]]`, we might look into using geojson-pydantic. I'm considering introducing pydantic for query validation eventually.

https://developmentseed.org/geojson-pydantic/intro/
https://github.com/developmentseed/geojson-pydantic

PR #315, opened 2026-05-19
Add datetime handling to WaterData Client

jarq6c:
This PR adds datetime handling to `waterdata_client`. Prior to this PR, client `get` methods with the `datetime` argument only accepted `str` values (or `Optional[str]` typically). These arguments can now accept `datetime.datetime`, `datetime.timedelta`, `pandas.Timestamp`, and `pandas.Timedelta`. These objects are transformed into `str` RFC 3339 datetime expressions prior to URL generation. `datetime` arguments still accept `str` values, which are passed directly to the API.

This functionality supports use-cases where date ranges may be derived from Python objects (e.g. `datetime.datetime.now`, `pandas.Series.dt.date`, etc.). This allows users to supply *datetime-like* objects directly without having to transform them into API-compatible strings. The derived datetime strings are also normalized to UTC to avoid a common complication when working with national-scale data.

## Example

```python
from datetime import datetime
from hydrotools.waterdata_client import ContinuousClient

client = ContinuousClient()
data = client.get(
    monitoring_location_id="USGS-02146470",
    datetime=datetime(2026, 5, 15, 1, 30)
)
```

## How does this work?

The datetime-handling systems consists of five type aliases and two functions. The steps are as follows:

1. Every request model with a `datetime` attribute has its type hint replaced with a new `TypeAlias` called `DateTimeField`. This type is `Annotated` with a Pydantic `BeforeValidator` that calls `validate_datetime_argument` on instantiation, but before Pydantic built-in schema validation.
2. `validate_datetime_argument` either returns a `str` input as-is, or dispatches datetime conversion to `convert_datetime_to_string`.
3. `convert_datetime_to_string` takes a `DateTimeInput` (which could be a `DateTimeObject`, `DurationObject`, or `TimeInterval`) and returns a correctly formatted API-compatible string.
4. The resulting string from steps 2 or 3 is set as the actual field in the Pydantic request model.
5. An error during string conversion will ultimately raise a Pydantic `ValidationError`.

## Changes
 - `README.md` Correct minor grammatical mistakes and added datetime examples.
 - `build_clients.py` Added module level constants to support type overrides. In this case, the upstream OGC API JSON schema type is replaced.
 - `build_request_models.py` Added module level constants to support type overrides. In this case, the upstream OGC API JSON schema type is replaced.
 - `schema_extract.py` Add new functionality to inject type overrides and associated imports.
 - `client.py.j2` Updated template to add required imports for type overrides.
 - `request_model.py.j2` Updated template to add required imports for type overrides.
 - `clients/*.py` Regenerated all clients with new template.
 - `request_models/*.py` Regenerated all Pydantic request models with new template.

## Additions
 - `datetime_utilities.py` New module with methods to handle conversion of datetime objects to string and Pydantic validation.

## Tests
 - `test_clients.py` New datetime functionality tested on an actual client class.
 - `test_datetime_query.py` Tests various datetime object combinations on Pydantic request models. Includes testing for correctly raised `ValidationError`.
 - `test_datetime_conversion.py` Tests various datetime object combinations for the conversion method.

## Checklist

### DIFF
--- a/python/waterdata_client/src/hydrotools/waterdata_client/utilities.py
+++ b/python/waterdata_client/src/hydrotools/waterdata_client/utilities.py
@@ -1,0 +1,99 @@
+"""Generally useful methods.
+"""
+from datetime import datetime, timedelta, UTC
+from typing import TypeAlias, Union, Literal, Sequence
+import pandas as pd
+
+DateTimeLike: TypeAlias = Union[datetime, pd.Timestamp]
+"""Type alias for date+time objects."""
+
+DurationLike: TypeAlias = Union[timedelta, pd.Timedelta, Sequence[DateTimeLike | None]]
+"""Type alias for time duration objects."""
+
+def convert_datetime_to_string(
+        datetime_object: DateTimeLike | DurationLike,
+        timespec: Literal['auto', 'hours', 'minutes', 'seconds', 'milliseconds', 'microseconds', 'nanoseconds'] = "seconds"
+) -> str:
+    """Converts a datetime-like or duration-like object to a standardized RFC 3339
+    string in UTC.
+    
+    Args:
+        datetime_object: A datetime, timedelta, or sequence of datetime compatible
+            objects to convert (e.g. [start, end]).
+        timespec: Maximum precision of datetime strings. Defaults to "seconds".
+    
+    Returns:
+        A date-time or interval string, adhering to RFC 3339 format. Intervals may be 
+        bounded or half-bounded (double-dots at start or end). Duration objects
+        like timedelta or pandas.Timedelta are converted to an equivalent bounded
+        interval ending at the current time.
+
+    Example:
+    >>> # Single datetime object conversion
+    >>> convert_datetime_to_string(datetime(2025, 1, 3))
+    ... "2025-01-03T00:00:00+00:00"
+
+    >>> # Timedelta conversion, assuming current time is 2025-01-03T00:00:00+00:00
+    >>> convert_datetime_to_string(timedelta(hours=36))
+    ... "2025-01-01T12:00:00+00:00/2025-01-03T00:00:00+00:00"
+
+    >>> # Sequence of datetimes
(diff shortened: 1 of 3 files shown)


python/waterdata_client/scripts/build_request_models.py
```diff
@@ -25,6 +25,18 @@
...
 
+TYPE_OVERRIDES: dict[str, str] = {
+    "datetime": "DateTimeField"
+}
+"""Mapping from specific arguments to special type hints. This facilitates
+various query interventions.
+"""
+
+EXTRA_IMPORTS: dict[str, str] = {
+    "datetime": "from ..datetime_utilities import DateTimeField"
```
christophertubbs: Enormous landmine. This won't explode _now_, but you're placing a hard to diagnose issue in the future. Add a ticket for it so remind you to find a solution that doesn't result in relative imports.

python/waterdata_client/src/hydrotools/waterdata_client/clients/__init__.py
```diff
@@ -5,7 +5,7 @@
 
 Package version: 0.8.0a0
 Generation script: build_clients.py
-Generated: 2026-05-13 17:38:51 Z
+Generated: 2026-05-18 18:31:08 Z
```
christophertubbs: Add a ticket that will _only_ spit out this update if the content of the data, not metadata, changes.

christophertubbs (changes requested):
This is a fine addition, though there are two identifiable points of added tech debt. Add tickets for these and I'll approve.

python/waterdata_client/scripts/build_request_models.py
jarq6c: I went ahead and replaced all the relative imports with explicit imports.

python/waterdata_client/src/hydrotools/waterdata_client/clients/__init__.py
jarq6c: Added. See https://github.com/NOAA-OWP/hydrotools/issues/316

